PR #30, opened 2026-02-25
ESM3 Encoder + Misc Refactors

vratins:
PR to add ESM3 encoder support and refactor embedding pipeline                                                                                                                                                            

- Added new ESMEncoder that reads pre-computed ESM3 embeddings as an alternative to GVP/SLAE encoders
- Added scripts/generate_esm_embeddings.py to precompute and cache ESM3 residue-level embeddings with proper biotite alignment
- Added scripts/generate_slae_embeddings.py for SLAE embedding generation
- Refactor dataset module to support ESM/SLAE cached embeddings and improve per-water quality filtering
- Update training script with pluggable encoder architecture (--encoder_type gvp|slae|esm) 
- GVP encoder with configurable message/update GVP layers
- Added comprehensive docstrings throughout codebase
- Updated READme for usage instructions


scripts/generate_esm_embeddings.py
```diff
@@ -0,0 +1,363 @@
...
+    Parse PDB with biotite to get ground truth residue list.
+
+    Uses the same residue counting method as geometry cache:
+    unique (chain_id, res_id, ins_code) tuples in sorted order.
+
+    Returns:
+        biotite_seq: List of single-letter codes (X for non-canonical)
+        num_residues: Total residue count
+    """
+    pdb_file = PDBFile.read(pdb_path)
```
marcuscollins: Is there a particular reason to use PDB instead of mmCIF/pdbx? mmCIF is considerably more robust. If you are using PDB, please add an issue to migrate to CIF.

scripts/generate_esm_embeddings.py
```diff
@@ -0,0 +1,363 @@
...
+
+    Returns:
+        aligned_emb: (N_biotite, embed_dim) aligned embeddings
+    """
+    num_biotite = len(biotite_seq)
+    embed_dim = esm_emb.shape[1]
+    aligned_emb = torch.zeros(num_biotite, embed_dim, dtype=esm_emb.dtype)
+
+    esm_idx = 0
+    for bt_idx, bt_aa in enumerate(biotite_seq):
```
marcuscollins: If you have the ism sequence, a good validation as you iterate would be to check that the res names are the same when you copy an embedding vector into the aligned_emb tensor.

scripts/generate_esm_embeddings.py
```diff
@@ -0,0 +1,363 @@
...
+        if output.embeddings is None:
+            raise RuntimeError("Model returned no embeddings")
+
+        emb = output.embeddings
+        if emb.dim() == 3:
+            emb = emb.squeeze(0)
+        emb = emb[1:-1].detach().cpu()  # remove BOS/EOS
+
+        # Remove chain breaks from ESM embeddings
+        esm_seq_with_breaks = protein.sequence or ""
```
marcuscollins: If the sequence is empty, can you actually proceed? Should this raise an error?

scripts/generate_slae_embeddings.py
```diff
@@ -0,0 +1,530 @@
...
+
+import argparse
+import sys
+from collections.abc import Iterable, Iterator
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent.parent))
+
+# SLAE paths - derive from home directory for portability
+SLAE_ROOT = Path.home() / "SLAE_wl"
```
marcuscollins: Why not have the user install the SLAE package into their site-packages?

scripts/generate_slae_embeddings.py
```diff
@@ -0,0 +1,530 @@
...
+from SLAE.io.atom_tensor import atom37_to_atoms, atomarray_to_tensors  # noqa: E402
+from SLAE.model.encoder import ProteinEncoder  # noqa: E402
+from SLAE.util.constants import PROTEIN_ATOMS  # noqa: E402
+from torch_geometric.data import Batch, Data  # noqa: E402
+from tqdm import tqdm  # noqa: E402
+
+from src.utils import setup_logging_for_tqdm  # noqa: E402
+
+
+def _normalize_ins_code(value) -> str:
```
marcuscollins: This is a repeated function. You should put it in some utilities module.

scripts/generate_slae_embeddings.py
```diff
@@ -0,0 +1,530 @@
...
+                continue
+
+            parts = line.split("_")
+            if len(parts) < 2:
+                logger.warning(f"Skipping malformed line: {line}")
+                continue
+
+            pdb_id = parts[0]
+            if parts[-1] != "final":
+                logger.warning(f"Unexpected format: {line}")
```
marcuscollins: What does this mean? This is a sort of hidden requirement that we should try to avoid. We have this problem in Sampleworks too, that we're parsing metadata from directory structures, which ultimately brittle. I would find a way to avoid this if possible and check what you actually mean to check (is it something about the structure or contents of the corresponding pdb file?)

scripts/generate_slae_embeddings.py
```diff
@@ -0,0 +1,530 @@
...
+                    "pdb_id": pdb_id,
+                    "pdb_path": pdb_path,
+                    "cache_key": line,
+                }
+            )
+
+    return entries
+
+
+def parse_pdb_protein_atoms(pdb_path: str | Path) -> bts.AtomArray:
```
marcuscollins: Repeated?

scripts/generate_slae_embeddings.py
```diff
@@ -0,0 +1,530 @@
...
+    protein_atoms = atoms[protein_mask]
+
+    return protein_atoms
+
+
+def get_geometry_atom_info(
+    protein_atoms: bts.AtomArray,
+) -> list[tuple[str, int, str, str]]:
+    """
+    Build list of (chain_id, res_id, ins_code, atom_name) for each atom in geometry order.
```
marcuscollins: what is "geometry order"?

scripts/generate_slae_embeddings.py
```diff
@@ -0,0 +1,530 @@
...
+        List of (chain_id, res_id, ins_code, atom_name) tuples, one per atom
+    """
+    return [
+        (
+            protein_atoms.chain_id[i],
+            protein_atoms.res_id[i],
+            _normalize_ins_code(protein_atoms.ins_code[i]),
+            protein_atoms.atom_name[i].strip().upper(),
+        )
+        for i in range(len(protein_atoms))
```
marcuscollins: This is probably a good place to use `zip`, and then you can also use `strict=True` to make sure the lists are all the same length. You can use numpy's np.str operations I think to vectorize things like upper-casing.

scripts/generate_slae_embeddings.py
```diff
@@ -0,0 +1,530 @@
...
+    Returns:
+        aligned_emb: (N_geometry, 128) embeddings aligned to geometry order
+    """
+    n_geometry = len(geometry_atom_info)
+    embed_dim = slae_emb.shape[1]
+    aligned = torch.zeros(n_geometry, embed_dim, dtype=slae_emb.dtype)
+
+    # Build lookup: (chain_id, res_id, ins_code, atom_name) -> slae_atom_index
+    slae_lookup = {}
+    for slae_idx in range(slae_emb.shape[0]):
```
marcuscollins: I feel like there should be a way to vectorize this and avoid the loop. It would be more compact and faster. Please make an issue for it and add a TODO comment with a link to that issue here.

scripts/generate_esm_embeddings.py
```diff
@@ -0,0 +1,363 @@
...
+Generate and cache ESM3 residue-level embeddings for protein structures.
+
+This script:
+1. Reads a split file containing PDB entries
+2. For each entry, loads the PDB with biotite (ground truth residue list)
+3. Runs ESM3 to get per-residue embeddings
+4. Aligns ESM embeddings to biotite structure (inserting zeros for skipped residues)
+5. Saves embeddings to separate .pt files in cache_dir/esm/
+
+ESM3's PDB parser skips certain non-canonical amino acids (SEC, PYL) entirely,
```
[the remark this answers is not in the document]
vratins: ESM offers residue level embeddings so I do not think we need to worry about alt-locs

scripts/generate_esm_embeddings.py
```diff
@@ -0,0 +1,363 @@
...
+        default=None,
+        help="Limit number of files to process (for testing)",
+    )
+    parser.add_argument(
+        "--overwrite", action="store_true", help="Overwrite existing embeddings"
+    )
+
+    args = parser.parse_args()
+
+    base_pdb_dir = Path(args.base_pdb_dir)
```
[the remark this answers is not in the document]
vratins: Addressed in latest PR

scripts/generate_slae_embeddings.py
vratins: Addressed in latest PR

scripts/generate_slae_embeddings.py
```diff
@@ -0,0 +1,530 @@
...
+        res_id = slae_residue_ids[res_idx].item()
+        ins = _normalize_ins_code(slae_ins_codes[res_idx])
+        atom_type_idx = slae_atom_type[slae_idx].item()
+        atom_name = PROTEIN_ATOMS[atom_type_idx]
+        slae_lookup[(chain, res_id, ins, atom_name)] = slae_idx
+
+    # Map geometry atoms to SLAE atoms
+    for geom_idx, key in enumerate(geometry_atom_info):
+        if key in slae_lookup:
+            aligned[geom_idx] = slae_emb[slae_lookup[key]]
```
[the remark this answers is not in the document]
vratins: Addressed in latest PR

Issue #27, opened 2026-02-11
Add pre-commit checks


marcuscollins:
Add a pre-commit check file to run ruff (format and check) and pyright or mypy before committing code. This will help catch a lot of issues.

PR #36, opened 2026-02-26
Constants + Utils updates

vratins:
Small PR that adds graph creation constants to `constants.py` and makes changes to `utils.py` to use those constants (as well slightly improving the tqdm logging utility function.

Also makes formatting changes to util plotting functions.

### DIFF
--- a/src/constants.py
+++ b/src/constants.py
@@ -1,16 +1,18 @@
 # constants.py
 """
 Constants for edge types and other shared values.
-
-Using constants instead of string literals helps prevent typos and enables
-IDE autocompletion and refactoring support.
 """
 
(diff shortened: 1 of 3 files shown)


DorisMai:
Util updates are primarily stylistic changes, so merging shouldn’t harm the main branch. My personal preference is that the visualization-related functions might be cleaner as a separate script. They are not part of the core model logic and have quite a lot of hard-coded args like color and marker size that might be subject to quite some changes in the future. But this is really just a minor/non-blocking comment.

DorisMai (approved):
Style changes (organizing import order) should hopefully be addressed after merging PR #39 and #45 where pre-commit hooks run to check and fix these things (address issue #27 ). Approving for now to address the current batch of PRs.

PR #37, opened 2026-02-26
ESM Encoder Class Creation

vratins:
- Add ESMEncoder for pre-computed ESM3 embeddings
- Update encoder interface to return edge features as third output
- Add use_edge_update parameter to GVPEncoder for optional edge updates
- Use constants for defaults
- Update tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -231,6 +231,8 @@ warn_return_any = true
 warn_unused_configs = true
 
 [tool.ty.rules]
+# uvx runs ty in isolation without project dependencies - ignore missing imports
+unresolved-import = "ignore"
 # PyTorch Geometric has incomplete type stubs - dynamic attributes on Batch/Data
 unresolved-attribute = "ignore"
 # PyG's MessagePassing.message() override pattern is intentional


src/encoder_base.py
DorisMai: Do we expect most encoders work by loading cached scalar embeddings? If so, the base class could be more exact with the `__init__` and `output_dim` functions and save some boiler plates in `esm/slae_encoder.py`.  It looks like so far `gvp_encoder` is the only outlier case which can just overload all these functions.

vratins: Added a class that separates encoders into cached embedding classes and boilerplate encoders (if in the future we decide to train our own encoder or something along those line)

src/gvp_encoder.py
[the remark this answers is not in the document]
vratins: I added this to the `GVPEncoder` base class -- made more sense there. In the future, when we add ligands, I intend on removing this method altogether since we would operate with `HeteroData` end to end.

src/gvp_encoder.py
```diff
@@ -411,6 +521,7 @@ def from_config(cls, config: dict, device: torch.device) -> GVPEncoder:
                 node_scalar_in=node_scalar_in,
                 hidden_dims=(hidden_s, hidden_v),
                 edge_scalar_in=16,
+                use_edge_update=use_edge_update,
```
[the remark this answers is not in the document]
vratins: Done

src/encoder_base.py
```diff
@@ -29,28 +29,32 @@ class BaseProteinEncoder(ABC, nn.Module):
     @abstractmethod
     def output_dims(self) -> tuple[int, int]:
         """Return (scalar_dim, vector_dim) output dimensions."""
-        pass
+        raise NotImplementedError("Subclasses must implement output_dims")
```
marcuscollins: Nit/fun fact: IIRC marking something as `@abstractmethod` will cause this error to be raised anyway

src/encoder_base.py
```diff
@@ -38,17 +38,21 @@ def encoder_type(self) -> str:
         pass
 
     @abstractmethod
-    def forward(self, data: HeteroData) -> tuple[torch.Tensor, torch.Tensor]:
+    def forward(
+        self, data: HeteroData
+    ) -> tuple[torch.Tensor, torch.Tensor, tuple | None]:
```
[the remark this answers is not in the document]
marcuscollins: In fact, I would say you need to keep the signature of this method fixed: if one of them returns three things, all cousins of this method should return the same three things. Otherwise you aren't properly abstracting the behavior.

src/encoder_base.py
```diff
@@ -65,13 +69,60 @@ def from_config(cls, config: dict, device: torch.device) -> BaseProteinEncoder:
...
+        Args:
+            data: HeteroData with cached embeddings in data['protein']
+
+        Returns:
+            s: (N, embedding_dim) — raw embeddings
+            V: (N, 0, 3)         — empty vector features
+            pp_edge_attr: None   — cached embedding encoders don't process edges
+        """
+        if self._embedding_key not in data['protein']:
+            raise NotImplementedError(
```
marcuscollins: This is a KeyError, not a NotImplementedError

src/gvp_encoder.py
```diff
@@ -99,9 +52,37 @@ def __init__(
...
     ):
+        """
+        Initialize GVP encoder for protein structure processing.
+
+        Args:
+            node_scalar_in: Input scalar feature dimension (e.g., element one-hot)
+            node_vec_in: Input vector feature channels (typically 1 for orientation)
+            hidden_dims: (scalar_dim, vector_dim) hidden layer dimensions
+            edge_scalar_in: Input edge scalar dimension (RBF features)
+            edge_vec_in: Input edge vector channels (unit vectors)
```
marcuscollins: Please document what shape should this have. It's a set of unit vectors, or a count? If it's a count, be consistent with others here and use a name like "n_edge_vec_in".

src/gvp_encoder.py
```diff
@@ -99,9 +52,37 @@ def __init__(
...
+            n_feedforward: Number of GVPs in feedforward function
+            drop_rate: Dropout rate for regularization
+            vector_gate: Whether to use vector gating in GVP layers
+            scalar_activation: Activation function for scalar channels
+            vector_activation: Activation function for vector gating
+            init_vec_zero: If True, initialize input vectors as zeros
+            pooled_dim: Output dimension when pooling by residue
+            pool_residue: If True, pool atom features to residue level
+            pool_aggr: Aggregation method for residue pooling ('mean', 'sum', 'max')
+            update_w_distance: Include distance features in edge updates
```
marcuscollins: again, use names that are a little more clear, e.g., `update_w_distance_features`.

src/gvp_encoder.py
```diff
@@ -178,36 +163,108 @@ def __init__(
 
     @staticmethod
     def _tuple_to_scalar_dense(x_tuple: tuple) -> torch.Tensor:
+        """
+        Convert GVP tuple to dense scalar representation.
```
marcuscollins: This should document that the method also normalizes V before concatenation to s.

src/gvp_encoder.py
```diff
@@ -178,36 +163,108 @@ def __init__(
...
+        x_scalar: torch.Tensor, device: torch.device | None = None
+    ) -> tuple[torch.Tensor, torch.Tensor]:
         zeros = torch.zeros(x_scalar.size(0), 1, 3, device=x_scalar.device if device is None else device)
         return (x_scalar, zeros)
 
-    def _compute_edge_attr(self, pos: torch.Tensor, edge_index: torch.Tensor):
-        d, u = _edge_vectors(pos, edge_index)
-        s_edge_raw = _rbf(d, num_gaussians=self.num_edge_rbf, cutoff=self.radius)
+    @staticmethod
+    def _edge_vectors(pos: torch.Tensor, edge_index: torch.Tensor):
```
marcuscollins: why move this into the class? To guide you, make something a `@staticmethod` of a class when the logic is specific to that class, even if it doesn't use class variables. Otherwise, leave it as its own method, to promote re-use where reasonable.

src/gvp_encoder.py
```diff
@@ -178,36 +163,108 @@ def __init__(
...
         V_edge = u.unsqueeze(1)
         return (s_edge, V_edge), s_edge_raw
 
-    def forward(self, data: Batch):
+    def forward(self, data: Batch) -> tuple[tuple, tuple | None]:
+        """
+        Forward pass through the GVP encoder.
+
+        Args:
+            data: Batch with node features, positions, and edge indices
```
marcuscollins: It is helpful to the reader to specify what the keys or attributes actually are. The best way to do this is to  define a data class and just say the data object must be of class x and if there are optional attributes for the class that are required here, specify them.

src/encoder_base.py
```diff
@@ -65,13 +69,60 @@ def from_config(cls, config: dict, device: torch.device) -> BaseProteinEncoder:
...
+                f"{self.encoder_type.upper()} encoder requires cached embeddings. "
+                f"Please provide pre-computed {self._embedding_key} in data['protein']."
+            )
+
+        embeddings = data['protein'][self._embedding_key]
+        V = embeddings.new_empty(embeddings.size(0), 0, 3)
```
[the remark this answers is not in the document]
vratins: Done

src/gvp_encoder.py
vratins: Added more detail

src/encoder_base.py
```diff
@@ -13,9 +15,71 @@
...
+        config: Configuration dictionary containing:
+            - encoder_type: 'gvp', 'slae', or 'esm' (required)
+            - Other encoder-specific parameters
+        device: Device to place the encoder on
+
+    Returns:
+        Instantiated encoder implementing BaseProteinEncoder
+    """
+    if "encoder_type" not in config:
+        raise ValueError("'encoder_type' must be specified in config")
```
marcuscollins: This is a KeyError

tests/test_flow.py
```diff
@@ -92,177 +90,180 @@ def mock_forward(data):
...
+        src = torch.tensor(
+            [[0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [2.0, 0.0, 0.0]], device=device
+        )
+        dst = torch.tensor([[0.5, 0.0, 0.0], [1.5, 0.0, 0.0]], device=device)
+
         edges = build_knn_edges(src, dst, k=2)
-        
+
         assert edges.shape[0] == 2
         assert edges.shape[1] > 0
```
marcuscollins: Presumably you can test exactly which nodes the edges connect--that would be a stronger test.

tests/test_encoder.py
```diff
@@ -167,43 +183,58 @@ def test_gvp_implements_interface(self, device, sample_hetero_data):
...
+        s, V, pp_edge_attr = encoder(sample_hetero_data_with_slae)
+        assert s.shape[0] == sample_hetero_data_with_slae["protein"].num_nodes
+        assert s.shape[1] == 128
+        assert V.shape == (sample_hetero_data_with_slae["protein"].num_nodes, 0, 3)
+        # Cached embedding encoder should return None for edge features
+        assert pp_edge_attr is None
+
+        # output_dims available after forward
         assert isinstance(encoder.output_dims, tuple)
         assert len(encoder.output_dims) == 2
```
marcuscollins: This and the previous line are redundant with the next assertion.

tests/test_encoder.py
```diff
@@ -322,78 +370,175 @@ def test_wrapper_freeze(self, device):
...
     def test_encoder_no_learnable_params(self, device):
-        """SLAE encoder should have no learnable parameters."""
-        encoder = SLAEEncoder(slae_dim=128).to(device)
+        """Cached embedding encoder should have no learnable parameters."""
+        encoder = CachedEmbeddingEncoder(
+            embedding_key="slae_embedding", encoder_type="slae"
+        ).to(device)
         assert sum(p.numel() for p in encoder.parameters()) == 0
 
+    def test_slae_from_config(self, device, sample_hetero_data_with_slae):
```
marcuscollins: pretty sure this is redundant.

src/encoder_base.py
vratins: Done

tests/test_encoder.py
vratins: Done

tests/test_encoder.py
```diff
@@ -225,28 +256,38 @@ def test_encoder_initialization(self, simple_encoder):
...
 
-        output = encoder(sample_homogeneous_data)
+        (s, v), edge_attr = encoder(sample_homogeneous_data)
 
-        assert isinstance(output, tuple)
-        s, v = output
         assert s.shape == (sample_homogeneous_data.num_nodes, 64)
         assert v.shape == (sample_homogeneous_data.num_nodes, 16, 3)
+        # edge_attr should be a tuple (s_edge, V_edge)
+        assert edge_attr is not None
```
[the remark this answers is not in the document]
vratins: Added clarification (`edge_atrr` would be `None` in all cases unless edge updates are enabled)

tests/test_encoder.py
```diff
@@ -283,36 +326,41 @@ def test_wrapper_encoder_type(self, device):
...
             hidden_dims=(64, 16),
-            edge_scalar_in=16,
+            n_edge_scalar_in=16,
             pool_residue=False,
         ).to(device)
 
         encoder = GVPEncoder(encoder=base_encoder, freeze=False)
 
-        s, V = encoder(sample_hetero_data)
+        s, V, edge_attr = encoder(sample_hetero_data)
```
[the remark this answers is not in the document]
vratins: Added issue

PR #42, opened 2026-03-01
Minimal refactor tests

DorisMai:
This PR changes the `tests` to prepare for PR #39 that sets up build workflow and helps clean up dependencies.

Changes are minimal. Functionally, the main change is replacing a hard-coded data file path with a relative one. This involves moving the `test_files` folder inside `test`. A `conftest.py` script is added to reduce some boiler plates on fixture.


tests/conftest.py
```diff
@@ -0,0 +1,59 @@
...
+TEST_DIR = Path(__file__).parent
+
+
+@pytest.fixture
+def device():
+    return torch.device("cuda" if torch.cuda.is_available() else "cpu")
+
+
+@pytest.fixture
+def pdb_base_dir():
```
marcuscollins: Does this need to be a fixture? It will certainly work, but it will call os.environ.get every time you want to get what is essentially a constant.

tests/conftest.py
```diff
@@ -0,0 +1,59 @@
...
+    env = os.environ.get("TEST_DATA_DIR")
+    if env:
+        return Path(env)
+    return TEST_DIR / "test_files"
+
+
+@pytest.fixture
+def pdb_path(pdb_base_dir):
+    """Factory fixture that resolves a PDB path, skipping if missing."""
+    def _resolve(pdb_id: str) -> str:
```
marcuscollins: Unless you need pdb_base_dir to be a fixture, I would simplify this by creating a constant that is generated once at initialization, e.g. `PDB_BASE_DIR` and then use that in this method. Then you don't need the method to return a function.

marcuscollins (approved):
A couple minor things to think about for the future maybe, otherwise LGTM.

vratins (approved):
LGTM

PR #70, opened 2026-03-24
Hydra Configs

### LINKED ISSUE #13: Remove hard coded paths everywhere in the codebase

### LINKED ISSUE #29: Configuration files for training runs

### LINKED ISSUE #32: Rename `<atom_type>_x` to `<atom_type>_feat` for clarity

vratins:
PR towards:
- Replacing `argparse` with hydra; defaults stored in `configs`
- Replace all hard-coded paths with relative paths or no default
- Replace `_x` with `_feat` for all nodes (trivial change hence in this PR) 

## Summary by CodeRabbit

* **Chores**
  * Migrated training and inference CLIs to Hydra; scripts now use structured configs and require explicit data directory inputs.
  * Updated runtime deps to include Hydra/OmegaConf and adjusted visualization package.

* **Documentation**
  * README overhauled for Hydra usage; CLI flag docs replaced with configuration structure and key mappings.
  * Added comprehensive YAML configs for data, flow, models, logging, training, and inference.

* **Bug Fixes**
  * Standardized cached feature naming and updated cache handling.

* **Tests**
  * Tests updated to use structured configs (OmegaConf) instead of argparse.


configs/data/default.yaml
```diff
@@ -0,0 +1,25 @@
+# Data configuration
+
+processed_dir: null
+base_pdb_dir: null
```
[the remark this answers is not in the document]
marcuscollins: @vratins I would favor validation in train.py over requiring them here. One thing I don't like about Hydra is that it easily becomes opaque, so that you don't know how, and can't easily figure out, what the required argument needs to be. I don't know of a good way to _document_ options when using Hydra configs. One way is to make sure to validate all the parameters up front in your scripts and provide useful error messages, although frankly that's pretty tedious.

configs/inference.yaml
```diff
@@ -0,0 +1,39 @@
...
+pdb_list: ???
+output_dir: ???
+
+# Data directories
+processed_dir: null
+base_pdb_dir: null
+
+# Inference settings
+inference:
+  include_mates: false
```
marcuscollins, line 18: Shouldn't this (and similar options) depend on how the model was trained? Perhaps you need to inherit from the training config?

scripts/inference.py
```diff
@@ -7,24 +7,20 @@
...
 Usage:
-    python scripts/inference.py \
-        --run_dir /path/to/run_directory \
-        --pdb_list /path/to/pdb_list.txt \
-        --output_dir /path/to/output \
-        --method rk4 \
-        --num_steps 100 \
-        --save_gifs
+    python scripts/inference.py run_dir=... pdb_list=... output_dir=...
+    python scripts/inference.py run_dir=... pdb_list=... output_dir=... integration.method=euler
```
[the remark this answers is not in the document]
marcuscollins: yeah, definitely need to fix this.

scripts/inference.py
```diff
@@ -44,166 +40,30 @@
...
 
-    with open(config_path, "r") as f:
-        config = json.load(f)
+    from typing import cast
+
+    from omegaconf import OmegaConf
+
+    config = cast(
+        dict, OmegaConf.to_container(OmegaConf.load(config_path), resolve=True)
+    )
```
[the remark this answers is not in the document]
marcuscollins: I think this is going to mean that you need to go through the rest of the code and make sure wherever these options are used, you are looking at the correct part of the configuration. I haven't looked at the unit tests yet, but these should thoroughly look at the config and make sure it is loaded and _used_ correctly.

scripts/inference.py
```diff
@@ -399,22 +259,24 @@ def save_plot(
...
-    run_dir = Path(args.run_dir)
-    output_root = Path(args.output_dir)
+    run_dir = Path(cfg.run_dir)
+    output_root = Path(cfg.output_dir)
     output_dir = output_root / run_dir.name
     output_dir.mkdir(parents=True, exist_ok=True)
     (output_dir / "plots").mkdir(exist_ok=True)
-    if args.save_gifs:
+    if cfg.inference.visualization.save_gifs:
         (output_dir / "gifs").mkdir(exist_ok=True)
```
marcuscollins, line 274: Why gifs?

scripts/train.py
```diff
@@ -10,24 +10,21 @@
...
 
 Usage:
-    python scripts/train.py \\
-        --train_list /path/to/train.txt \\
-        --val_list /path/to/val.txt \\
-        --encoder_type gvp \\
-        --epochs 200 \\
-        --batch_size 4
+    python scripts/train.py train_list=splits/train.txt val_list=splits/val.txt
+    python scripts/train.py train_list=... model=slae training.batch_size=8
```
marcuscollins, line 14: You might need to document all the options here.

scripts/train.py
```diff
@@ -46,357 +43,110 @@
...
+def _extract_water_filter_config(cfg: DictConfig) -> dict:
     """Extract per-water filtering parameters (toggleable)."""
     return {
-        "max_protein_dist": args.max_protein_dist,
-        "min_edia": args.min_edia,
-        "max_bfactor_zscore": args.max_bfactor_zscore,
-        "filter_by_distance": args.filter_by_distance,
-        "filter_by_edia": args.filter_by_edia,
-        "filter_by_bfactor": args.filter_by_bfactor,
+        "edia_dir": cfg.data.edia_dir,
```
marcuscollins, line 69: In general try to avoid introducing new options when making a big PR like this that changes config systems. Changing the config system at the same time as config options just makes it harder to review.

scripts/train.py
```diff
@@ -46,357 +43,110 @@
...
+        "max_protein_dist": cfg.data.water_filter.max_protein_dist,
+        "min_edia": cfg.data.water_filter.min_edia,
+        "max_bfactor_zscore": cfg.data.water_filter.max_bfactor_zscore,
+        "filter_by_distance": cfg.data.water_filter.filter_by_distance,
+        "filter_by_edia": cfg.data.water_filter.filter_by_edia,
+        "filter_by_bfactor": cfg.data.water_filter.filter_by_bfactor,
     }
```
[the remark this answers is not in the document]
marcuscollins: Have you run a test training run to make sure this works?

scripts/train.py
```diff
@@ -562,15 +312,17 @@ def run_eval_sampling(
 
         out = flow_matcher.rk4_integrate(
             graph,
-            num_steps=args.rk4_steps,
-            use_sc=args.use_self_cond,
+            num_steps=cfg.logging.rk4_steps,
```
marcuscollins, line 315: What is this num_steps? If the option is for logging you should rename it `num_logging_steps`, otherwise it seems like it is a number of integration steps.

scripts/train.py
```diff
@@ -648,20 +400,21 @@ def train_epoch(
 
     pbar = tqdm(train_loader, desc=f"Epoch {epoch} [Train]")
     for step, batch in enumerate(pbar):
-        batch = batch.to(args.device)
+        batch = batch.to(cfg.logging.device)
```
marcuscollins, line 403: It feels weird for device to be under the logging part of the configuration. I think this will lead to confusion.

scripts/train.py
```diff
@@ -648,20 +400,21 @@ def train_epoch(
...
-            use_self_conditioning=args.use_self_cond,
-            accumulation_steps=args.grad_accum_steps,
+            use_self_conditioning=cfg.flow.use_self_cond,
+            accumulation_steps=cfg.training.optimizer.grad_accum_steps,
         )
 
         if metrics["per_sample_info"] is not None:
-            per_sample_losses = metrics["per_sample_info"]["losses"].cpu()
-            num_graphs = metrics["per_sample_info"]["num_graphs"]
+            per_sample_info = cast(dict, metrics["per_sample_info"])
```
marcuscollins, line 415: Why introduce this `cast` statement?

configs/logging/default.yaml
```diff
@@ -0,0 +1,18 @@
...
+save_every: 10
+eval_every: 5
+n_eval_samples: 3
+rk4_steps: 100
+save_gifs: false
+threshold: 1.0
+
+log_level: INFO
+log_file: null
+device: cuda
```
marcuscollins, line 14: Weird place for this.

scripts/train.py
```diff
@@ -849,90 +605,107 @@ def build_scheduler(optimizer, args):
...
+        duplicate_single_sample=cfg.data.duplicate_single_sample,
         **dataset_kwargs,
     )
 
     # sample fixed eval indices
     np.random.seed(42)
     eval_indices = np.random.choice(
         len(val_loader.dataset),
-        min(args.n_eval_samples, len(val_loader.dataset)),
+        min(cfg.logging.n_eval_samples, len(val_loader.dataset)),
```
marcuscollins, line 708: I hadn't noticed this before--why do you limit the number of eval samples?

scripts/train.py
```diff
@@ -951,43 +724,40 @@ def main():
     node_scalar_in = int(sample_data["protein"].x.shape[-1])
     logger.info(f"Detected protein input dimension: {node_scalar_in}")
 
-    log_encoder_sample_stats(sample_data, args.encoder_type)
+    log_encoder_sample_stats(sample_data, cfg.model.encoder_type)
     encoder_config = resolve_encoder_config(
```
marcuscollins, line 728: is this what is needed for inference? Should you log it and/or save it in the config file you save above, and to wandb?

scripts/train.py
```diff
@@ -951,43 +724,40 @@ def main():
...
-        dir=args.wandb_dir,
-        name=args.run_name,
+        project=cfg.logging.wandb.project,
+        dir=cfg.logging.wandb.dir,
+        name=run_name,
         config=config_dict,
     )
 
-    model = build_model(args, device, encoder_config=encoder_config)
+    model = build_model(cfg, device, encoder_config=encoder_config)
```
marcuscollins, line 752: Isn't the `encoder_config` derived just from `cfg` and maybe a couple other parameters? Could you just create the encoder config inside `build_model`.

tests/test_train_config.py
```diff
@@ -61,16 +59,20 @@ def test_resolve_embedding_dim_raises_on_embedding_type_mismatch(
...
-        embedding_dim=None,
+    cfg = OmegaConf.create(
+        {
+            "model": {
+                "encoder_type": "slae",
+                "hidden_s": 256,
+                "hidden_v": 64,
+                "freeze_encoder": False,
+                "encoder_ckpt": None,
+                "embedding_dim": None,
```
marcuscollins, line 70: do you not need the rest of the parameters in here?

tests/test_train_config.py
```diff
@@ -1,15 +1,13 @@
-from argparse import Namespace
-
 import pytest
 import torch
+from omegaconf import OmegaConf
```
marcuscollins, line 3: I strongly suggest doing a lot more testing here, in particular testing the pattern of defaults and making sure things change the way you expect. As the AI reviewers above noted, it's easy in a config system that has a hierarchy of defaults to specify and/or read things incorrectly. I would make sure the values you expect come through the way you expect.

README.md
```diff
@@ -218,66 +225,75 @@ To resume training from a checkpoint, you can load the model weights and optimiz
...
+| Config Key | Default | Description |
+|------------|---------|-------------|
+| `train_list` | required | Path to training split file |
+| `val_list` | required | Path to validation split file |
+| `model` | `gvp` | Model config: `gvp`, `esm`, or `slae` |
+| `training.batch_size` | `4` | Batch size (use smaller for ESM) |
+| `training.optimizer.grad_accum_steps` | `1` | Gradient accumulation steps |
+| `training.epochs` | `200` | Number of training epochs |
+| `training.optimizer.lr` | `1e-3` | Learning rate |
+| `training.scheduler.type` | `cosine` | LR scheduler: `cosine`, `step`, `none` |
```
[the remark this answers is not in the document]
marcuscollins: Definitely fix.

marcuscollins (changes requested):
Mainly I want to make sure this code actually runs--from the Copilot comments it seems like it won't, but the fact that you've been training models suggest it will run fine. There's a third, worse possibility--that it will run, but with opaque defaults. So having more config tests should be included, to make sure the model is created the way you expect.

PR #82, opened 2026-06-17
KNN edge construction fix + align residue indexing with ESM sanitization

vratins:
* `build_knn_edges` was calling `knn(x=dst_pos, y=src_pos)`, which queries each `src` point's nearest `dst` points instead of each `dst` point's nearest `src` points; swapped the call and the resulting index rows to fix this. This was previously masked by taking the union of edges on both sides. A future PR will use KNN as a fallback to radius graphs. Expanded the docstring to document that the query is per-destination, so every destination is guaranteed incoming edges (`row 1`) while a source that is nobody's nearest neighbor may be absent from `row 0` — and updated the water-water coverage tests to assert on the destination row (`row 1`) accordingly, removing the now-unnecessary `xfail` markers.
* `match_atoms_to_coords` now also handles an empty `atoms` array instead of only an empty `target_coords` array.
* Residue indices for the protein graph are now computed via the same residue-name canonicalization the ESM embedding generation script uses (`THREE_TO_ONE` -> `ONE_TO_THREE`, unknowns -> `UNK`) before counting residue boundaries with biotite's `get_residue_starts`. Without this, two residues that share (`chain`, `resid`, `ins_code`) but had different original `res_names` could get merged into one under ESM's sanitization but stay separate here, desyncing residue counts/indices from the stored ESM embeddings. Insertion codes are now also normalized (`normalize_ins_code`) before counting, since `get_residue_starts` splits on `ins_code` too — a blank vs. placeholder code (`''` / `'.'` / `'?'`) would otherwise split or merge residues differently from the ESM script's residue keys. Both the canonicalization (`sanitize_res_names_for_esm`) and the insertion-code normalization are now shared between `src/dataset.py` and `scripts/generate_esm_embeddings.py` to prevent the two paths from drifting apart.
* **Cleanup:** Removed a dead duplicate `build_knn_edges` in `src/utils.py` (the canonical one lives in `src/flow.py`) that still carried the old per-source semantics, and removed the orphaned `atom37_to_atoms` / `ATOM37_FILL` helper (unused outside its own tests; the SLAE pipeline uses its own implementation). Added regression tests covering residue-count alignment between the dataset and the ESM residue keys.

## Summary by CodeRabbit

- **Improvements**
  - Improved protein residue indexing to better align residue counts and atom-to-residue expansion with ESM-style conventions, including insertion-code handling.
  - Enhanced residue-name normalization for ESM embeddings and dataset indexing, canonicalizing modified/unknown residues consistently.
  - Improved water-position dataset documentation to match current prediction behavior.
  - Corrected directed KNN/water edge orientation and strengthened destination-row coverage checks to ensure consistent graph connectivity.
  - Added clearer warnings when atom-to-coordinate matching is substantially incomplete.

- **Dependencies**
  - Pinned minimum versions for PyTorch Geometric packages.

### DIFF
--- a/src/dataset.py
+++ b/src/dataset.py
@@ -656,7 +656,7 @@ def filter_waters_by_quality(
 
 class ProteinWaterDataset(Dataset):
     """
-    Dataset for protein crystal contact prediction.
+    Dataset for predicting water positions in protein crystal structures.
 
     Returns HeteroData with:
     - 'protein' node type: ASU protein atoms + optionally symmetry mates
@@ -991,21 +991,13 @@ class ProteinWaterDataset(Dataset):
         # both -> UNK at the same position). Apply the same renaming here so the
         # residue count and indices align with the stored ESM embeddings.
         sanitized_for_idx = protein_atoms.copy()
-        for _i in range(len(sanitized_for_idx)):
-            _aa1 = THREE_TO_ONE.get(sanitized_for_idx.res_name[_i], "X")
-            sanitized_for_idx.res_name[_i] = ONE_TO_THREE.get(_aa1, "UNK")
+        for i in range(len(sanitized_for_idx)):
+            aa1 = THREE_TO_ONE.get(sanitized_for_idx.res_name[i], "X")
+            sanitized_for_idx.res_name[i] = ONE_TO_THREE.get(aa1, "UNK")
         res_starts = bts.get_residue_starts(sanitized_for_idx)
         num_residues = len(res_starts)
-        protein_res_idx_np = np.zeros(len(protein_atoms), dtype=np.int64)
-        for res_i in range(len(res_starts)):
-            start = res_starts[res_i]
-            end = (
-                res_starts[res_i + 1]
-                if res_i + 1 < len(res_starts)
-                else len(protein_atoms)
-            )
-            protein_res_idx_np[start:end] = res_i
-        protein_res_idx = torch.tensor(protein_res_idx_np, dtype=torch.long)
+        atom_res_idx = np.searchsorted(res_starts, np.arange(len(protein_atoms)), side="right") - 1
+        protein_res_idx = torch.from_numpy(atom_res_idx.astype(np.int64))
         num_waters = len(water_atoms)
         ratio_valid, ratio_reason = check_water_residue_ratio(
             num_waters,
--- a/src/dataset.py
+++ b/src/dataset.py
@@ -29,7 +29,14 @@ from torch_cluster import radius_graph
 from torch_geometric.data import Batch, HeteroData
 from tqdm import tqdm
 
-from src.constants import EDGE_PP, ELEM_IDX, ELEMENT_VOCAB, NUM_RBF
+from src.constants import (
+    EDGE_PP,
+    ELEM_IDX,
+    ELEMENT_VOCAB,
+    NUM_RBF,
+    ONE_TO_THREE,
+    THREE_TO_ONE,
+)
 from src.utils import (
     compute_edge_features,
     normalize_ins_code,
@@ -140,7 +147,7 @@ def match_atoms_to_coords(
     Returns:
         List of indices into atoms array for matched atoms
     """
-    if target_coords.shape[0] == 0:
+    if target_coords.shape[0] == 0 or len(atoms) == 0:
         return []
 
     matched = []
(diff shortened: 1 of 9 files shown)


tests/test_flow.py
```diff
@@ -595,9 +598,11 @@ def test_batched_waters_have_edges(self, batched_hetero_data):
             f"Only {len(water_nodes_with_pw_edges)}/{n_water} waters have protein edges in batched data"
         )
 
-        # Check water-water edges
+        # Check water-water edges. WW edges are built per destination, so every
+        # water appears as a destination (row 1); assert coverage on the
+        # destination/query row rather than the source row.
         if n_water > 1:
-            water_nodes_with_ww_edges = torch.unique(ww_edges[0])
+            water_nodes_with_ww_edges = torch.unique(ww_edges[1])
```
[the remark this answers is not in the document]
vratins: Done.

src/dataset.py
DorisMai, line 162: This is a comment outside the diff as I look at this function. Should probably add a check (and log warning if failed) that the length of atoms and target_coords should roughly match, and majority atoms matched.

src/utils.py
```diff
@@ -117,6 +82,36 @@ def normalize_ins_code(value) -> str:
     return ins
 
 
+def sanitize_res_names_for_esm(atoms):
```
DorisMai: Add type hint on input and output

src/dataset.py
```diff
@@ -977,20 +983,22 @@ def _preprocess_one(self, entry: dict, cache_path: Path):
...
+        # embeddings.
+        sanitized_for_idx = sanitize_res_names_for_esm(protein_atoms)
+        for i in range(len(sanitized_for_idx)):
+            sanitized_for_idx.ins_code[i] = normalize_ins_code(
+                sanitized_for_idx.ins_code[i]
+            )
+        res_starts = bts.get_residue_starts(sanitized_for_idx)
+        num_residues = len(res_starts)
+        atom_res_idx = (
+            np.searchsorted(res_starts, np.arange(len(protein_atoms)), side="right") - 1
```
DorisMai: maybe name the variables clearer or add a comment explaining why you are not using atomarray.res_id. The same sanitization can probably be achieved by `bts.spread_residue_wise(sanitized_for_idx, np.arange(num_res))` but as long as it's correct I don't particularly care how it is implemented. With that, probably worth adding a test checking you are assigning the res_id correctly.

tests/test_utils.py
```diff
@@ -132,80 +133,85 @@ def test_normalize_valid_code(self):
...
+                    element="C",
+                )
+                for (res_id, res_name, ins) in residues
+            ]
+        )
 
-        assert torch.allclose(coords[0], expected_coord)
+    @staticmethod
+    def _esm_key_count(atoms):
+        """Replicate the ESM script's residue-key counting."""
```
DorisMai, line 169: if truly a replicate then why not extract this part of ESM script into a helper function to import here?

src/flow.py
```diff
@@ -34,12 +34,37 @@ def build_knn_edges(
...
     """
     if src_pos.numel() == 0 or dst_pos.numel() == 0:
         return torch.empty(2, 0, dtype=torch.long, device=src_pos.device)
 
-    idx = knn(x=dst_pos, y=src_pos, k=k, batch_x=batch_dst, batch_y=batch_src)
+    # knn(x=src_pos, y=dst_pos) returns row 0 = dst (query) indices, row 1 = src
+    # (neighbor) indices; swap so the result follows the src(row 0)->dst(row 1)
+    # edge_index convention.
+    idx = knn(x=src_pos, y=dst_pos, k=k, batch_x=batch_src, batch_y=batch_dst)
+    idx = torch.stack((idx[1], idx[0]), dim=0)
```
DorisMai, line 209: Please add an exact set asymmetric test with toy cases where you know true neighbors to test on the directionalities, both within the `knn` function (to explicitly fail the `x` vs `y` mix up) and for interpreting the output tensor. specifically for the output, I couldn't find anywhere officially documented that the knn output tensor row 0 is destination indices and row 1 is source indices. moreover, it looks like the latests `knn` function in torch_geometry is different from the 2.7.0 version this repo currently uses, so it might worth making a note here in addition to explicitly test the behavior stays the same.

tests/test_flow.py
DorisMai, line 947: this is just a curiosity question outside the diff. It looks like PW is the only type with symmetrized knn edges built. Is there a particular motivation for this?

DorisMai (changes requested):
The only blocking request I have is add a test for the knn fix to loudly fail the mix up (previous bug) and guards any additional mix up the ordering. All other comments are nitpick.

tests/test_flow.py
vratins: This was initially done to ensure coverage in the P<->W case, the other edge types had coverage by default. Not an issue since in future PRs we switch completely to radius graphs with a kNN fallback.

src/utils.py
vratins: Done.

src/flow.py
vratins: Tried this and you were right the existing test couldn't catch it. I reverted `build_knn_edges` to the buggy x/y order and test_basic_knn still passed, because its geometry has distance ties. New `TestBuildKnnEdgesDirection` uses lopsided geometry (correct → {(0,0),(0,1)}, swapped → {(0,0),(1,1),(2,1)}); 2 of its 4 tests fail on that revert. I verified the row order empirically on the pinned 2.7.0 (row 0 = y query, row 1 = x neighbor), pinned in a test. Added a note for the versioning.

src/dataset.py
vratins: Good catch, was not aware of this method. Replaced it here since it seems cleaner since I do not need to use `searchsorted`.

tests/test_utils.py
vratins: Fair point, but that logic is scheduled for deletion in the next PR that is built on top of this PR. The ESM script is moving to biotite's residue segmentation which drops the tuple-key counting entirely and makes the two sides agree by construction. A shared helper now is one we'd delete in that change.

src/flow.py
vratins: Also upgraded torch-geometric to 2.8 and pyg-lib to 0.6 so we're on the latest

src/dataset.py
vratins: Fixed this. ALso updated the docstring to be more clear about what the method is doing.

PR #83, opened 2026-06-17
Adding CIF file support

### LINKED ISSUE #33: Migrate to CIF files

vratins:
`dataset.py` and `utils.py` now support reading `.cif` files alongside `.pdb`:

-  `_read_structure()` dispatches to biotite's CIF or PDB reader based on file extension, used by `parse_asu_with_biotite` and compute_normalized_bfactors.

-  `compute_normalized_bfactors` no longer re-reads the file when bfactors are already available from `parse_asu_with_biotite`; the shared logic moves to `_compute_normalized_bfactors_from_atoms()`.

- ` _parse_pdb_list` now prefers a PDB ID's `.cif` file over `.pdb` when both exist (checked via a single `os.path.isfile`, no directory scan).

- `utils.resolve_structure_path()` resolves a preferred CIF/PDB path to whichever file actually exists on disk; `parse_split_file()` now returns preferred `.cif` paths for callers to resolve lazily.

## Summary by CodeRabbit

* **New Features**
  * Added support for CIF structure files alongside PDB.
  * When both are available, CIF is preferred for dataset and split-file structure resolution.
* **Bug Fixes**
  * More robust handling of missing/unreadable structure files during embedding generation and preprocessing, with more consistent failure behavior.
* **Improvements**
  * Faster B-factor normalization by reusing already-parsed structure data and ensuring required atom fields are available.
  * Improved crystal-contact handling to use the resolved structure source.
* **Tests**
  * Expanded CIF parsing, structure-selection, and dataset/cache behavior coverage.

### DIFF
--- a/src/dataset.py
+++ b/src/dataset.py
@@ -62,7 +62,7 @@ def _read_structure(path: str | Path, extra_fields=None) -> bts.AtomArray:
 
 
 def parse_asu_with_biotite(
-    path: str,
+    path: str | Path,
 ) -> tuple[bts.AtomArray, bts.AtomArray]:
     """
     Parse PDB or CIF file and extract protein and water atoms.
@@ -96,7 +96,7 @@ def parse_asu_with_biotite(
 
 
 def get_crystal_contacts_pymol(
-    pdb_path: str, cutoff: float = 5.0
+    struc_path: str, cutoff: float = 5.0
 ) -> dict[str, np.ndarray | list]:
     """
     Extract ASU and symmetry mate atoms within crystal contact distance.
@@ -105,7 +105,7 @@ def get_crystal_contacts_pymol(
     interface atoms within the specified cutoff distance.
 
     Args:
-        pdb_path: Path to PDB file with crystal symmetry information
+        struc_path: Path to structure file (PDB/CIF) with crystal symmetry information
         cutoff: Distance cutoff in Angstroms for interface detection
 
     Returns:
@@ -120,7 +120,7 @@ def get_crystal_contacts_pymol(
         cmd.reinitialize()
         cmd.feedback("disable", "all", "everything")
         obj = "struct"
-        cmd.load(pdb_path, obj)
+        cmd.load(struc_path, obj)
         cmd.symexp("sym", obj, obj, cutoff)
         cmd.select("interface", f"byres (sym* within {cutoff} of {obj})")
 
@@ -492,7 +492,7 @@ def load_edia_for_pdb(
 
 
 def compute_normalized_bfactors(
-    pdb_path: str,
+    struc_path: str,
 ) -> tuple[dict[tuple[str, int, str], float] | None, np.ndarray | None]:
     """
     Extract and normalize B-factors for water molecules.
@@ -501,7 +501,7 @@ def compute_normalized_bfactors(
     in the selected structure.
 
     Args:
-        pdb_path: Path to PDB file
+        struc_path: Path to structure file (PDB/CIF)
 
     Returns:
         Tuple of:
@@ -510,7 +510,7 @@ def compute_normalized_bfactors(
         Returns (None, None) on error
     """
     try:
-        atoms = _read_structure(pdb_path, extra_fields=["b_factor"])
+        atoms = _read_structure(struc_path, extra_fields=["b_factor"])
 
         # filter for water molecules
         water_mask = (atoms.res_name == "HOH") | (atoms.res_name == "WAT")
@@ -519,7 +519,7 @@ def compute_normalized_bfactors(
         return _compute_normalized_bfactors_from_atoms(water_atoms)
 
     except Exception as e:
-        logger.warning(f"Warning: Could not extract B-factors from {pdb_path}: {e}")
+        logger.warning(f"Warning: Could not extract B-factors from {struc_path}: {e}")
         return None, None
 
 
@@ -822,7 +822,7 @@ class ProteinWaterDataset(Dataset):
         for pdb_id, cache_key in pdb_ids:
             subdir = self.base_pdb_dir / pdb_id
             cif_path = subdir / f"{pdb_id}_final.cif"
-            pdb_path = (
+            struc_path = (
                 cif_path if os.path.isfile(cif_path) else subdir / f"{pdb_id}_final.pdb"
             )
 
@@ -830,7 +830,7 @@ class ProteinWaterDataset(Dataset):
             entries.append(
                 {
                     "pdb_id": pdb_id,
-                    "pdb_path": pdb_path,
+                    "struc_path": struc_path,
                     "cache_key": cache_key,
                     "embedding_key": cache_key,  # Same as cache_key for embedding lookup
                 }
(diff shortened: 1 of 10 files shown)


src/dataset.py
```diff
@@ -776,31 +796,42 @@ def _parse_pdb_list(self, pdb_list_file: str) -> list[dict]:
         Constructs path: {base_pdb_dir}/{pdb_id}/{pdb_id}_final.pdb
```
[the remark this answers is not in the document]
vratins: Addressing in next commit.

src/utils.py
```diff
@@ -129,7 +147,9 @@ def parse_split_file(split_file: Path, base_pdb_dir: Path) -> list[dict]:
         base_pdb_dir: Base directory containing PDB subdirectories
 
     Returns:
-        List of entry dicts with keys: pdb_id, pdb_path, cache_key
+        List of entry dicts with keys: pdb_id, pdb_path, cache_key.
+        `pdb_path` is a preferred `.cif` path; callers should resolve the
+        actual existing file with `resolve_structure_path()` when parsing.
```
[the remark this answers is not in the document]
vratins: Addressing in next commit.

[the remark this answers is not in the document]
vratins: Addressed

src/dataset.py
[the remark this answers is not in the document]
vratins: Addressed

src/utils.py
```diff
@@ -117,9 +117,27 @@ def normalize_ins_code(value) -> str:
     return ins
 
 
+def resolve_structure_path(path: str | Path) -> Path | None:
```
marcuscollins: This function seems a little superfluous. I don't think you need to look for a different file extension if the specified one isn't there. Just raise an error.

src/utils.py
marcuscollins: You can actually iterate over the file directly by using `for line in open(split_file, "r"):`, it will manage the file closing for you.

marcuscollins: Simplify this: `parts = line.strip().split("_")`

marcuscollins: That said, you should just create a csv or tsv file with the pieces you need. E.g., lines like:

1VME, /path/to/1VME.cif, <<cache_key>>

and then read them in with pandas or something similar.

tests/test_dataset.py
```diff
@@ -506,6 +507,58 @@ def test_water_residue_names(self, pdb_6eey):
...
+        """Should return protein and water atom arrays from CIF."""
+        protein_atoms, water_atoms = parse_asu_with_biotite(cif_6eey)
+
+        assert protein_atoms is not None
+        assert water_atoms is not None
+        assert len(protein_atoms) > 0
+
+    def test_cif_hydrogen_removed(self, cif_6eey):
+        """Hydrogens should be removed from CIF-parsed arrays."""
+        protein_atoms, water_atoms = parse_asu_with_biotite(cif_6eey)
```
marcuscollins: These could be a test fixtures since they are re-used many times. It would speed up your tests.

tests/test_dataset.py
```diff
@@ -868,6 +928,104 @@ def test_empty_lines_ignored(self, tmp_path, pdb_base_dir):
 
         assert len(dataset.entries) == 1
 
+    def test_prefers_cif_when_both_formats_exist(self, tmp_path):
+        """Split parsing should store the preferred CIF path."""
+        base_dir = tmp_path / "pdbs"
+        self._write_structure(base_dir, "abcd", [".cif", ".pdb"])
+
+        list_file = tmp_path / "list.txt"
```
marcuscollins, line 1136: can you be sure that the tmp directory is cleaned up?

tests/test_dataset.py
```diff
@@ -868,6 +928,104 @@ def test_empty_lines_ignored(self, tmp_path, pdb_base_dir):
...
+
+        assert [entry["pdb_id"] for entry in dataset.entries] == ["keep1", "keep2"]
+        assert dataset.entries[0]["pdb_path"] == (
+            base_dir / "keep1" / "keep1_final.pdb"
+        )
+        assert dataset.entries[1]["pdb_path"] == (
+            base_dir / "keep2" / "keep2_final.cif"
+        )
+
+    def test_parse_does_not_probe_filesystem(self, tmp_path, monkeypatch):
```
marcuscollins: This is a weird test. Is there some reason we care about this? Remove it if there isn't some very specific concern.

tests/test_dataset.py
```diff
@@ -868,6 +928,104 @@ def test_empty_lines_ignored(self, tmp_path, pdb_base_dir):
...
+
+        dataset = ProteinWaterDataset(
+            pdb_list_file=str(list_file),
+            processed_dir=str(tmp_path / "processed"),
+            base_pdb_dir=str(base_dir),
+            preprocess=False,
+        )
+
+        assert [entry["pdb_id"] for entry in dataset.entries] == ["keep1", "keep2"]
+        assert dataset.entries[0]["pdb_path"] == (
```
marcuscollins: This is testing a couple different things:
1. which ids are retained and which are discarded
2. a specific set of paths and file name conventions. 

In general a unit test should really test one thing. To achieve that here, you could create a method that generates the file paths from the pdb id, and then test that method separately.

tests/test_utils.py
```diff
@@ -131,6 +135,58 @@ def test_normalize_valid_code(self):
         assert normalize_ins_code(" B ") == "B"
 
 
+@pytest.mark.unit
+class TestStructurePathResolution:
+    """Tests for lazy CIF/PDB path resolution helpers."""
+
+    @staticmethod
+    def _write_structure(base_dir: Path, pdb_id: str, suffixes: list[str]) -> None:
```
marcuscollins: A lot of these tests are redundant with some in `test_dataset.py`. That just adds time to your test runs. For instance, a lot of the tests above are implicitly testing `resolve_structure_path`. Please go through these and minimize them. You can use a test coverage report to determine whether you've decreased your coverage when you remove any tests.

marcuscollins (commented):
There are several things that could be improved, particularly in the tests. Please have a look and see if you can make some changes.

src/utils.py
vratins: Done, created a single `stuc_path` that should be a cif or pdb, and errors out in the parse function if the file does not exist. Made this change wherever a "`pdb_path`" is invoked.

src/utils.py
vratins: Made the change for the loop. As for the second comment, do you mean my input split files should be formatted that way? Since the path and cache keys can change depending on the directory we specify, hence I chose to go with a simple text file, and we can specify where to find the structure files (path dir) and where to create the cache.

vratins: Done

tests/test_dataset.py
vratins: Yes, `pytest` `tmp_path` are cleaned up by default

tests/test_dataset.py
vratins: Done.

tests/test_utils.py
vratins: Addressing both this and the above comment, I've tried to remove some redundant tests, and consolidate as much as I can while minimizing coverage.

tests/test_dataset.py
vratins: Addressed below.

tests/test_dataset.py
vratins: Done

src/utils.py
```diff
@@ -129,38 +129,46 @@ def parse_split_file(split_file: Path, base_pdb_dir: Path) -> list[dict]:
...
     entries = []
-    with open(split_file, "r") as f:
-        for line in f:
-            line = line.strip()
-            if not line:
-                continue
-
-            parts = line.split("_")
-            if len(parts) < 2:
+    for line in open(split_file, "r"):
```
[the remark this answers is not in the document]
vratins: @marcuscollins I'll defer a decision on this to you -- is this needed?

PR #84, opened 2026-06-17
Adding explicit sampling strategies

vratins:
- `sample_waters_uniform_ball` — places each water uniformly inside a ball of radius cutoff around a randomly chosen protein anchor atom, with volume correction for sampling uniformly within a ball of radius `r`.

- `sample_waters_scaled_gaussian` — `N(0, (sigma^2)*I)` where `sigma` is the per-graph std of protein coordinates. Equivalent to old behavior but explicit.

- Added the above to the `FlowMatcher` class

- Tests cover shape/count correctness, the empty-graph edge case, and a real-structure test against 6eey that verifies the cutoff guarantee holds on actual protein geometry with two graphs of different water counts batched together.

## Release Notes

* **New Features**
  * Added two per-graph water sampling strategies for initializing noisy water positions: uniform within a cutoff ball and scaled Gaussian with per-graph sigma.
  * Introduced selectable sampling and dynamic edge policy options, including automatic resolution based on batch conditions.
  * Updated inference/integration APIs to accept `water_count` (takes precedence over `water_ratio`), with consistent per-graph water node rebuilding.

* **Tests**
  * Added coverage for sampling correctness, cutoff guarantees, empty-water handling, deterministic behavior, policy auto-resolution, and negative `water_count` validation.


src/flow.py
```diff
@@ -486,6 +598,8 @@ def __init__(
         t_distort: float = 0.5,
         sigma_distort: float = 0.5,
         loss_eps: float = 1e-3,
+        sampling_strategy: str = "uniform_ball",
+        dynamic_edge_policy: str = "auto",
     ):
```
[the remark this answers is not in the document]
vratins: Not needed.

src/flow.py
```diff
@@ -26,6 +26,115 @@
...
+    num_p_per_graph = scatter(
+        torch.ones(batch_p.size(0), device=device, dtype=torch.long),
+        batch_p.to(device),
+        dim=0,
+        dim_size=num_graphs,
+        reduce="sum",
+    )
+    offsets = torch.zeros(num_graphs + 1, dtype=torch.long, device=device)
+    offsets[1:] = num_p_per_graph.cumsum(dim=0)
+
```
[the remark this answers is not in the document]
vratins: Addressed

src/flow.py
```diff
@@ -477,6 +586,9 @@ class FlowMatcher:
     High level class for flow matching training, validation, and numerical integration
     """
 
+    SAMPLING_STRATEGIES = {"uniform_ball", "scaled_gaussian"}
+    DYNAMIC_EDGE_POLICIES = {"auto", "radius", "knn_if_isolated"}
+
```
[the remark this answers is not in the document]
vratins: Addressed

src/flow.py
```diff
@@ -579,15 +753,22 @@ def training_step(
...
         x1 = batch["water"].pos
         batch_w = batch["water"].batch
         batch_p = batch["protein"].batch
-        num_graphs = int(batch_p.max().item()) + 1
-
-        sigma = self.compute_sigma(batch)
-
-        x0 = torch.randn_like(x1) * sigma
+        num_graphs = self._num_graphs(batch)
+
```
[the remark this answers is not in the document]
vratins: Addressed

src/flow.py
```diff
@@ -674,14 +850,21 @@ def validation_step(self, batch: HeteroData) -> dict[str, float]:
...
 
         x1 = batch["water"].pos
         batch_w = batch["water"].batch
         batch_p = batch["protein"].batch
-        num_graphs = int(batch_p.max().item()) + 1
-
-        sigma = self.compute_sigma(batch)
-        x0 = torch.randn_like(x1) * sigma
+        num_graphs = self._num_graphs(batch)
+
```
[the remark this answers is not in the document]
vratins: Addressed

src/flow.py
```diff
@@ -800,19 +1021,26 @@ def euler_integrate(
...
-            ) * sigma_per_water.unsqueeze(-1)
+            num_graphs = len(graphs)
+            num_waters = scatter(
+                torch.ones(batch_w.size(0), device=device, dtype=torch.long),
+                batch_w,
+                dim=0,
+                dim_size=num_graphs,
+                reduce="sum",
+            )
+            x, batch_w = self._sample_waters(g, num_waters, device)
```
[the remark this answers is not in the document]
vratins: Addressed

src/flow.py
```diff
@@ -912,17 +1142,26 @@ def rk4_integrate(
...
-            x = torch.randn_like(x1_true) * sigma_per_water.unsqueeze(-1)
+            num_graphs = len(graphs)
+            num_waters = scatter(
+                torch.ones(batch_w.size(0), device=device, dtype=torch.long),
+                batch_w,
+                dim=0,
+                dim_size=num_graphs,
+                reduce="sum",
+            )
+            x, batch_w = self._sample_waters(g, num_waters, device)
```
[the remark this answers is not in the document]
vratins: Addressed

tests/test_flow.py
```diff
@@ -452,6 +471,157 @@ def test_sample_rk4(self, flow_matcher, simple_hetero_data, device):
...
+
+        pos, batch_w = sample_waters_uniform_ball(
+            protein_pos=protein_pos, batch_p=batch_p,
+            num_waters=num_waters, cutoff=8.0, device=device,
+        )
+
+        assert pos.shape == (301, 3)
+        assert batch_w.shape == (301,)
+
+    def test_real_structure_cutoff_and_batch(self, device, pdb_6eey):
```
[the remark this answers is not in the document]
vratins: Addressed

src/flow.py
```diff
@@ -773,11 +993,12 @@ def euler_integrate(
...
             water_ratio: If provided, sample num_residues * water_ratio waters
                         instead of using ground truth water count
+            water_count: If provided, sample exactly this many waters per protein
 
         Returns:
             List of dicts, one per input graph, each with keys:
                 'protein_pos': (Np, 3) - includes both ASU and mate atoms
-                'water_true': (Nw, 3) - None if water_ratio is used
+                'water_true': (Nw, 3) - None if water_ratio/water_count is used
                 'water_pred': (Nw, 3) final prediction
```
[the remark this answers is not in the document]
vratins: Addressed

src/flow.py
```diff
@@ -885,11 +1114,12 @@ def rk4_integrate(
...
```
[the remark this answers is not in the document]
vratins: Addressed

src/flow.py
```diff
@@ -579,15 +753,22 @@ def training_step(
 
         self.model.train()
         device = batch["protein"].pos.device
+        batch.dynamic_edge_policy = self._effective_dynamic_edge_policy()
```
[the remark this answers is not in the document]
vratins: Left for future PR

src/flow.py
```diff
@@ -722,23 +919,58 @@ def _setup_water_nodes_from_ratio(
...
-        )
+        x, batch_w = self._sample_waters(g, num_waters, device)
         total_waters = batch_w.size(0)
 
-        # compute sigma per graph and expand to per-water
-        sigma_per_graph = self.compute_sigma_per_graph(g, device)
-        sigma_per_water = sigma_per_graph[batch_w]
+        # create water features (oxygen one-hot, index 2 for 'O' in ELEMENT_VOCAB)
+        water_x = torch.zeros(total_waters, 16, device=device)
+        water_x[:, 2] = 1.0  # oxygen is index 2 in ELEMENT_VOCAB
```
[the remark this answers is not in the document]
vratins: Only a concern if we change our element vocab

src/flow.py
```diff
@@ -26,6 +26,127 @@
...
+    )
+
+    # per-graph protein atom counts and cumulative offsets
+    num_p_per_graph = scatter(
+        torch.ones(batch_p.size(0), device=device, dtype=torch.long),
+        batch_p.to(device),
+        dim=0,
+        dim_size=num_graphs,
+        reduce="sum",
+    )
```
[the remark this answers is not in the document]
vratins: Addressed

src/flow.py
```diff
@@ -498,14 +630,74 @@ def __init__(
...
+                f"got '{dynamic_edge_policy}'"
+            )
         self.model = model
         self.p_self_cond = p_self_cond
         self.use_distortion = use_distortion
         self.p_distort = p_distort
         self.t_distort = t_distort
         self.sigma_distort = sigma_distort
         self.loss_eps = loss_eps
+        self.graph_cutoff = getattr(model, "cutoff", 8.0)
```
DorisMai, line 670: where do you ever set the attribute `cutoff` for `FlowWaterGVP` model? It looks like it is always just hardcoded to 8.0 here. Either declare a constant to use or add the attribute? and please use a clear name to separate it from the many other cutoff names already exist in the code base.

src/flow.py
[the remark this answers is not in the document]
DorisMai: which PR would this be (if not already opened)?

src/flow.py
```diff
@@ -624,35 +822,30 @@ def training_step(
...
         # check for high loss and compute per-sample losses for debugging
         per_sample_info = None
         if loss.item() > 100.0:
             with torch.no_grad():
                 from torch_scatter import scatter_add
 
                 weighted_mse = (w * per_atom_mse).squeeze(-1)
-                # compute per-graph loss: sum(weighted_mse) / sum(w) for each graph
                 numerator = scatter_add(weighted_mse, batch_w, dim=0)
                 denominator = scatter_add(w.squeeze(-1), batch_w, dim=0)
```
DorisMai: outside diff and you probably have guarded against no water somewhere else but maybe good to pass in `dim_size=num_graphs` in case of any empty graph?

src/flow.py
DorisMai, line 815: this part seems identical to `validation_step()` line 870-888? might be good to extract as _prepare_flow_batch

src/flow.py
```diff
@@ -624,35 +822,30 @@ def training_step(
         per_atom_mse = (v_pred - v_target).pow(2).mean(dim=-1, keepdim=True)
         loss = (w * per_atom_mse).sum() / w.sum()
 
+        # training RMSD
+        with torch.no_grad():
+            x1_hat = x_t + (1.0 - t_per_atom) * v_pred
+            diff2 = ((x1_hat - x1_star) ** 2).sum(-1)  # (Nw,)
+            rmsd = torch.sqrt(scatter_mean(diff2, batch_w, dim=0)).mean()
+
```
DorisMai, line 851: similarly shared code with `validation_step()` line 893-902, consider refactor to _flow_loss_and_rmsd

src/flow.py
DorisMai, line 826: outside diff but curious why self-conditioning is allowed during training but not during validation?

src/flow.py
```diff
@@ -722,23 +925,58 @@ def _setup_water_nodes_from_ratio(
...
+        Returns:
+            x: (N_water_total, 3) initial noise positions
+            batch_w: (N_water_total,) batch indices
+        """
+        if water_count < 0:
+            raise ValueError(f"water_count must be >= 0, got {water_count}")
+
+        num_residues = g["protein"].num_residues  # (num_graphs,)
+        num_graphs = num_residues.size(0)
```
DorisMai: why is num_graph computed differently here, rather than using `_num_graphs(batch)` like before?

src/flow.py
```diff
@@ -773,11 +1012,13 @@ def euler_integrate(
             device: Device to run on
             water_ratio: If provided, sample num_residues * water_ratio waters
                         instead of using ground truth water count
+            water_count: If provided, sample exactly this many waters per protein
```
DorisMai: should clarify the ordering of these parameters, which one overwrites the other (or would be ignored if the other is provided etc.)

src/flow.py
```diff
@@ -800,19 +1041,29 @@ def euler_integrate(
         x1_true = g["water"].pos.clone()
         batch_w_true = g["water"].batch.clone()
 
-        if water_ratio is not None:
```
DorisMai, line 803: similarly repeated in `rk4_integrate()` and could consider extract/refactor

src/flow.py
DorisMai: If you do decide to consistently use _num_graphs(batch), remember to change in `euler_integrate()` and `rk4_integrate()` too -- there without the inline comment to remind me this is about num_graph rather than num_residues can be confusing

tests/test_flow.py
```diff
@@ -452,6 +471,202 @@ def test_sample_rk4(self, flow_matcher, simple_hetero_data, device):
...
+            dists = torch.cdist(g_waters, g_protein)  # (n_w, n_atoms)
+            min_dists = dists.min(dim=1).values  # (n_w,)
+            assert min_dists.max().item() <= cutoff + 1e-4, (
+                f"Graph {g}: water too far from protein "
+                f"(max dist {min_dists.max().item():.4f} > {cutoff})"
+            )
+
+
+@pytest.mark.unit
+class TestScaledGaussianSampling:
```
DorisMai, line 642: should probably add a check that sigma of each graph is broadcasted correctly?

src/flow.py
DorisMai: outside diff but seems brittle to any changes on featurization. should avoid hard coded inline numbers as much as you can.

DorisMai (approved):
Ideally I would like the comments on computing num_graphs addressed (and if you plan on using sample_waters_scaled_gaussian at all in any experiment maybe add the test too). Otherwise mostly extraction/refactoring suggestions.

src/flow.py
vratins: Right so this probably leaked over from my changes with radius graphs, so I'll roll this back and add it with the PR that has radius graphs.

src/flow.py
vratins: Fair, switched to `_num_graphs(g)` throughout.

src/flow.py
vratins: Agreed, changed this to `len(ELEMENT_VOCAB) + 1` for the width and `ELEM_IDX["O"]` for the index

src/flow.py
vratins: `water_count` takes precedence over `water_ratio`, and when neither is given the ground-truth count is resampled from the prior; clarified.

src/flow.py
vratins: I've let it be here for now, but I plan on getting rid of it altogether. From further testing it seems that self-conditioning makes predictions worse and i will be removing it from our method altogether. and looking at literature it seems that it does best when designing or generating samples de-novo (like molecules) rather than conditioned on some other structure (like in our case).

src/flow.py
vratins: Added dim_size=num_graphs to both scatter_adds

src/flow.py
vratins: There's a lot of restructuring and feature addition to do here, so I'll consolidate once the methods settle.

src/flow.py
vratins: Will refactor in future PR.

tests/test_flow.py
vratins: Added `test_sigma_broadcasts_per_graph`

src/flow.py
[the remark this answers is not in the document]
vratins: This will be the PR opened once knn fix is merged, which will have radius graph construction over kNN, and using kNN as fall-back.

src/flow.py
vratins: I'll wrap and consolidate this in a PR where I implement AMP.

src/flow.py
```diff
@@ -795,24 +1098,13 @@ def euler_integrate(
         g = Batch.from_data_list([copy.deepcopy(graph) for graph in graphs]).to(device)
 
         batch_p = g["protein"].batch
+        num_graphs = self._num_graphs(g)
```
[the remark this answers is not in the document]
vratins: Left for next PR.

src/flow.py
```diff
@@ -907,22 +1205,13 @@ def rk4_integrate(
...
```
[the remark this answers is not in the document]
vratins: Left for next PR.

src/flow.py
```diff
@@ -498,14 +645,75 @@ def __init__(
...
+                device=device,
+            )
+        # scaled_gaussian
+        sigma_per_graph = self.compute_sigma_per_graph(batch_data, device)
+        return sample_waters_scaled_gaussian(
+            batch_w=batch_w,
+            sigma_per_graph=sigma_per_graph,
+            device=device,
+            dtype=batch_data["protein"].pos.dtype,
+        )
```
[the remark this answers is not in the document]
vratins: Done

src/flow.py
```diff
@@ -722,27 +935,63 @@ def _setup_water_nodes_from_ratio(
...
-        )
+        batch_w = _batch_from_counts(num_waters, device)
+        x = self._sample_waters(g, batch_w, device)
         total_waters = batch_w.size(0)
 
-        # compute sigma per graph and expand to per-water
-        sigma_per_graph = self.compute_sigma_per_graph(g, device)
-        sigma_per_water = sigma_per_graph[batch_w]
+        # create water features (oxygen one-hot; +1 for the trailing 'other' bucket)
+        water_x = torch.zeros(total_waters, len(ELEMENT_VOCAB) + 1, device=device)
```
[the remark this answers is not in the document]
vratins: Done

src/flow.py
```diff
@@ -722,27 +935,63 @@ def _setup_water_nodes_from_ratio(
...
+
+        batch_w = _batch_from_counts(num_waters, device)
+        x = self._sample_waters(g, batch_w, device)
+        total_waters = batch_w.size(0)
 
-        # create water features (oxygen one-hot, index 2 for 'O' in ELEMENT_VOCAB)
-        water_x = torch.zeros(total_waters, 16, device=device)
-        water_x[:, 2] = 1.0  # oxygen is index 2 in ELEMENT_VOCAB
+        # create water features (oxygen one-hot; +1 for the trailing 'other' bucket)
+        water_x = torch.zeros(total_waters, len(ELEMENT_VOCAB) + 1, device=device)
```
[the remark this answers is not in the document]
vratins: Done

PR #85, opened 2026-06-17
Optimizations to reduce per-step I/O overhead and to reduce graph density

vratins:
mmap-backed cache loading (`_load_torch_cache`)

- Wraps `torch.load` with `mmap=True` so `geometry/embedding .pt` files are memory-mapped. Falls back to regular load if the runtime or file format doesn't support it.
- Threaded through `load_slae_embedding`, `load_esm_embedding`, and `__getitem__`.
- Controlled by `--cache_load_mmap` (default: off).
  
Per-worker sample LRU cache (`sample_cache_size`)

- Optional in-process `OrderedDict`-backed LRU cache in `__getitem__`. When a sample is already in cache, skips all I/O and returns a `.clone()`. Evicts least-recently-used entries when the cache is full.
- 0 by default (disabled); set --`sample_cache_size` N to hold N samples per worker.
- default is 0, set it higher if enough available system RAM.
  
`max_neighbors` cap on `radius_graph`

- ProteinWaterDataset now accepts max_neighbors (default 256) and passes it to `radius_graph` at preprocessing time
- Stored in the geometry cache metadata for traceability.

## Release Notes

* **New Features**
  * Added `--sample_cache_size` command-line argument to enable per-worker in-process sample caching with configurable capacity
  * Added `--cache_load_mmap` flag to enable optimized dataset cache loading
  * Samples are cached in-process with automatic capacity management

* **Tests**
  * Added tests verifying cache data isolation between retrievals
  * Added tests confirming configuration propagation to dataset loaders

### DIFF
--- a/src/dataset.py
+++ b/src/dataset.py
@@ -698,7 +698,7 @@ class ProteinWaterDataset(Dataset):
         filter_by_edia: bool = True,
         filter_by_bfactor: bool = True,
         sample_cache_size: int = 0,
-        cache_load_mmap: bool = True,
+        cache_load_mmap: bool = False,
     ):
         """
         Args:
--- a/src/dataset.py
+++ b/src/dataset.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 
 import itertools
 import json
+from collections import OrderedDict
 from pathlib import Path
 
 import biotite.structure as bts
@@ -191,11 +192,24 @@ def _pad_atom_embeddings_for_mates(
     return torch.cat([asu_embedding, pad], dim=0)
 
 
+def _load_torch_cache(path: Path, *, cache_load_mmap: bool = True):
+    """Load a torch cache file, using mmap when supported by the file/runtime."""
+    if not cache_load_mmap:
+        return torch.load(path, weights_only=False)
+
+    try:
+        return torch.load(path, weights_only=False, mmap=True)
+    except (TypeError, ValueError, RuntimeError, OSError) as exc:
+        logger.debug(f"mmap torch.load failed for {path}; falling back: {exc}")
+        return torch.load(path, weights_only=False)
+
+
 def load_slae_embedding(
     embedding_dir: Path,
     cache_key: str,
     num_asu_protein: int,
     total_num_atoms: int,
+    cache_load_mmap: bool = True,
 ) -> torch.Tensor:
     """
     Load SLAE atom-level embeddings from cache.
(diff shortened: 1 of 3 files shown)


src/dataset.py
```diff
@@ -191,11 +192,24 @@ def _pad_atom_embeddings_for_mates(
     return torch.cat([asu_embedding, pad], dim=0)
 
 
+def _load_torch_cache(path: Path, cache_load_mmap: bool = True) -> dict:
+    """Load a torch cache file, using mmap when supported by the file/runtime."""
+    if not cache_load_mmap:
+        return torch.load(path, weights_only=False)
```
marcuscollins, line 198: should you pipe through `weights_only`?

src/dataset.py
```diff
@@ -745,6 +773,9 @@ def __init__(
         self.filter_by_distance = filter_by_distance
         self.filter_by_edia = filter_by_edia
         self.filter_by_bfactor = filter_by_bfactor
+        self.sample_cache_size = int(sample_cache_size)
+        self.cache_load_mmap = bool(cache_load_mmap)
+        self._sample_cache: OrderedDict[tuple[int, str], HeteroData] = OrderedDict()
```
marcuscollins, line 780: does it need to be ordered?

marcuscollins (approved):
A couple small things for you to consider before merging, but otherwise approving.

src/dataset.py
vratins: LRU semantics require it to be ordered; it uses `move_to_end` to mark recently-used keys and `popitem(last=False)` to evict the oldest, neither of which a regular `dict` supports.

src/dataset.py
vratins: the `weights_only` parameter is True by default, I've set it to False everywhere torch.load is used to suppress the pickle concern warning since we load in self-generated .pt cache files at every spot we use torch.load. I do not expect to set it to True anywhere, hence not threading it as a parameter.

src/dataset.py
```diff
@@ -717,14 +738,21 @@ def __init__(
...
+        if sample_cache_size < 0:
+            raise ValueError("sample_cache_size must be >= 0")
+
         self.cache_dir = Path(processed_dir)
         # Directory-based separation: geometry/ vs geometry_mates/
         cache_suffix = "_mates" if include_mates else ""
         self.geometry_dir = self.cache_dir / f"{geometry_cache_name}{cache_suffix}"
         self.base_pdb_dir = Path(base_pdb_dir)
         self.cutoff = cutoff
+        self.max_neighbors = max_neighbors
```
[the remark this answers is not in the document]
vratins: Addressed.

PR #88, opened 2026-08-04
Crystal-mates + cache meta-data

vratins:
- Cleans up the symmetry mates  code: per-atom and per-ligand dedup, `is_mate` `emb_res_idx` fields. Closes the special-position label leak, a water on a rotation axis is its own symmetry copy, i.e. the target sitting in the input.
 - Anchors the uniform-ball prior on ASU atoms only. No-mates runs unchanged.
 - Adds `_filter_meta.json` per cache dir; a run with different filter settings is refused instead of silently extending the cache.
 - Added tests for the above.

Default changes to match current checkpoint runs: `max_bfactor_zscore` 1.5 -> 2.0, `min_water_residue_ratio` 0.6 -> 0.1. Every cache used for training was built at the these values.

## Summary by CodeRabbit

* **New Features**
  * Improved handling of crystal symmetry mates, ligands, duplicate atoms, and embedding assignments.
  * Added cache metadata validation and filtering provenance tracking.
  * Improved water sampling around eligible protein atoms with fallback behavior when needed.

* **Bug Fixes**
  * Excluded mate waters from labeling and corrected ligand or unmatched-atom embeddings.
  * Updated quality-filter defaults to a 0.1 minimum water ratio and 2.0 B-factor z-score.

* **Documentation**
  * Expanded guidance on symmetry mates, masks, cache behavior, filtering, and inference.
  * Clarified that SLAE generation is retained for reproducibility while ESM is primarily used.


src/constants.py
DorisMai, line 1: why removing the comments?

README.md
```diff
@@ -114,20 +116,22 @@ Preprocessed data is cached under `--processed_dir` in a three-layer architectur
 │       - protein_pos: centered protein coordinates (N, 3)
 │       - protein_x: element one-hot encoding (N, 16)
 │       - protein_res_idx: residue indices for grouping
-│       - is_ligand: bool mask marking the appended ASU ligand atoms (N,)
+│       - is_ligand: bool mask marking the ligand atoms (N,)
+│       - is_mate: bool mask marking the symmetry-mate atoms (N,)
+│       - emb_res_idx: embedding row per atom; -1 means no row (N,)
```
DorisMai: you might want to rename `protein_pos/x/res_idx` to be more generic, maybe like `node_pos/x/res_idx` (referencing your line 124: `Node order is [ASU protein | mate protein | ASU ligand | mate ligand]`), since at this point they are not just protein atoms any more, and the array shape annotation `N` would get confusing.

src/dataset.py
```diff
@@ -914,6 +1058,76 @@ def _parse_pdb_list(self, pdb_list_file: str) -> list[dict]:
...
+
+        if write:
+            self.geometry_dir.mkdir(parents=True, exist_ok=True)
+            # Written through a temp file: cache builds fan out over processes,
+            # and a reader must never catch a half-written sidecar.
+            tmp_path = meta_path.with_suffix(f".{os.getpid()}.tmp")
+            with open(tmp_path, "w") as f:
+                json.dump(current, f, indent=2)
+            tmp_path.replace(meta_path)
+        elif any(self.geometry_dir.glob("*.pt")):
```
DorisMai: related to coderabbitai's comment. you probably want to move this `elif` branch up to be if the meta_path file does NOT exist, but `*.pt` exists, then log warning, regardless of whether write or not. This would change your `test_unlabelled_cache_warns`.

src/dataset.py
```diff
@@ -1100,6 +1333,20 @@ def _preprocess_one(self, entry: dict, cache_path: Path):
...
+        # (chain, res_id, ins_code) -> residue index, so a symmetry mate can
+        # inherit the embedding row of the ASU residue it is an image of. Keyed
+        # off the same sanitized parse that defines protein_res_idx, so the index
+        # lines up with the stored ESM rows.
+        asu_reskey_to_residx: dict[tuple[str, int, str], int] = {}
+        for res_i, start in enumerate(bts.get_residue_starts(sanitized_for_idx)):
+            key = (
+                str(sanitized_for_idx.chain_id[start]).strip(),
+                int(sanitized_for_idx.res_id[start]),
+                normalize_ins_code(sanitized_for_idx.ins_code[start]),
```
DorisMai: seems not needed, already normalized in line 1329

src/dataset.py
```diff
@@ -1118,33 +1365,69 @@ def _preprocess_one(self, entry: dict, cache_path: Path):
...
+                # A mate keeps its source residue's (chain, resi) and the embedding
+                # is coordinate-free, so the two share one row. -1 means no match,
+                # which reads as a zero embedding.
+                mate_emb_idx = []
+                for atom in mate_atoms:
+                    parsed = _parse_pdb_resi(atom.resi)
+                    mate_emb_idx.append(
+                        asu_reskey_to_residx.get((str(atom.chain).strip(), *parsed), -1)
+                        if parsed is not None
+                        else -1
```
DorisMai, line 1448: in what scenario would you expect to have a non-matched mate protein reskey? Should here be a warning or raise an error? And what does a zero-embedding of those extra atoms functionally do to the water prediction?

src/dataset.py
```diff
@@ -138,24 +159,39 @@ def get_crystal_contacts_pymol(
...
+            return coords if coords is not None else np.zeros((0, 3), dtype=float)
 
-        asu_coords = (
-            asu_coords if asu_coords is not None else np.zeros((0, 3), dtype=float)
-        )
-        mate_coords = (
-            mate_coords if mate_coords is not None else np.zeros((0, 3), dtype=float)
+        # Whole protein-mate residues with any atom within cutoff of the ASU.
+        cmd.select(
+            "iface_prot", f"byres ((sym* and polymer.protein) within {cutoff} of {obj})"
```
DorisMai: in addition to altlocs, this command might also give you Hydrogen atoms for mate protein (same thing for mate ligands)? unclear if it matters, but for your ASU atoms you did filter out Hydrogen.

src/dataset.py
```diff
@@ -1118,33 +1365,69 @@ def _preprocess_one(self, entry: dict, cache_path: Path):
...
+                mate_pos = torch.tensor(mate_coords, dtype=torch.float32) - center
+                mate_x = element_onehot([a.symbol.upper() for a in mate_atoms])
+
+                # compute mate residue indices (group atoms by actual residue)
+                mate_residue_keys = [(a.chain, a.resi) for a in mate_atoms]
+                unique_mate_res = list(dict.fromkeys(mate_residue_keys))  # keeps order
+                mate_res_map = {k: i for i, k in enumerate(unique_mate_res)}
+                mate_res_idx = torch.tensor(
+                    [mate_res_map[k] for k in mate_residue_keys], dtype=torch.long
+                )
```
DorisMai, line 1434: It is worth looking into coderabbitai's comment here for two reasons. 
1. `mate_res_idx` is later used to compute `final_protein_res_idx`, which is cached and loaded as `protein_res_idx` for computing `num_residues` (line 1650) which counts both asu and mate residues. Is this the same `num_residues` used to sample water at inference time? Is the intention that inference mode is always run without mates? otherwise, you might want to double check on the usage of `num_residues` vs `num_protein_residues` in the entire repo again, since these are older code outside the diff.
2. Currently you don't seem to do any pooling over residues, but if that's ever set True, then collision on (chain_id, res_id) for mate atoms on different symmetry copies would be a problem, and you would want to distinguish each symmetry mate protein separately.

src/flow.py
```diff
@@ -97,6 +102,14 @@ def sample_waters_uniform_ball(
     if batch_p.numel() > 0:
         num_graphs = max(num_graphs, int(batch_p.max().item()) + 1)
 
+    # Drop to the eligible anchors
+    if anchor_mask is not None:
+        eligible = anchor_mask.to(device).bool()
+        counts = torch.bincount(batch_p[eligible], minlength=num_graphs)
+        if not (counts[batch_w] == 0).any():
```
DorisMai: probably should warn in the else situation rather than quietly fall back to not use anchor_mask over the entire batch. Is it an intended behavior to drop mask for the entire batch? I thought you might only want to fall back for that particular graph. If so, you should probably make `test_anchor_mask_ignored_when_it_starves_a_graph` a stronger test where the second graph still has a mask used.

src/flow.py
```diff
@@ -57,6 +57,7 @@ def sample_waters_uniform_ball(
     batch_w: Tensor,
     cutoff: float = 8.0,
     device: torch.device | None = None,
+    anchor_mask: Tensor | None = None,
```
DorisMai, line 96: should you add `anchor_mask` for `sample_waters_scaled_gaussian` too? would mate protein affect the `sigma_per_graph computation`?

src/dataset.py
```diff
@@ -199,6 +233,114 @@ def match_atoms_to_coords(
...
+        earlier = [j for j in neighbors[i] if j < i and kept[j]]
+        # query_ball_point includes r, this sweep is strict.
+        dists = np.linalg.norm(mate_coords[earlier] - mate_coords[i], axis=1)
+        kept[i] = not (dists < tol).any()
+
+    keep_idx = np.flatnonzero(kept)
+    return mate_coords[keep_idx], [mate_atoms[i] for i in keep_idx]
+
+
+def dedup_mate_ligands_by_residue(
```
DorisMai, line 291: I am quite confused by the different treatment of mate protein vs ligand in the two `dedup_*` methods here. (1) Why are you not treating the mate protein in whole blocks (symmetry copy 1, copy 2, ...) like the whole ligand block? (2) Why does protein dedup needs first sweep + second round of neighbor look up, but ligand dedup only needs first sweep?

tests/test_dataset.py
```diff
@@ -305,6 +309,172 @@ def test_warns_on_odd_counts_below_half(self, warning_log, n_atoms, n_matched):
         assert f"{n_matched}/{n_atoms} atoms matched" in warning_log[0]
 
 
+@pytest.mark.unit
+class TestDedupMateAtoms:
```
DorisMai, line 313: I think it's important for `TestDedupMateAtoms` and `TestDedupMateLigandsByResidue` to have a real test case where you see dedup is really needed.

tests/test_dataset.py
DorisMai, line 1: Currently you don't have integration tests where both ligands and mates are turned on. The ligand integration tests all have `include_mates=False`. The mates tests uses 6eey which doesn't contain ligands. You should add a test class/case. Also, I think most integration tests here uses `encoder_type="gvp"`. If the paper uses esm, it's probably good to have at least one test here.

tests/test_dataset.py
```diff
@@ -2665,6 +2889,156 @@ def test_num_asu_protein_metadata_correct(
...
+        assert is_mate.shape == emb_res_idx.shape == (data["protein"].num_nodes,)
+        assert not is_mate[:num_asu].any()
+        assert is_mate.sum().item() > 0
+        # Ligands ride behind the mates, so the mate block is contiguous but need
+        # not run to the end; every marked node is past the ASU.
+        assert is_mate[num_asu:][: is_mate.sum().item()].all()
+
+        mate_protein = is_mate & ~cached["is_ligand"]
+        assert (emb_res_idx[mate_protein] >= 0).all()
+        assert (emb_res_idx[mate_protein] < data["protein"].num_protein_residues).all()
```
DorisMai: worth addressing the coderabbitai's comment  above.
Also, here, asserting just the range is quite weak. Since getting the `emb_res_idx` wrong was what prevented symmetry mates from being useful in the past, I think a stronger test on exact equivalence (for residues that you know are the same in ASU vs in mate) would be helpful.

DorisMai (changes requested):
I have some questions on the mates/ligands implementation. The connection between pymol handled mates/ligands and biotite handled ASU is in my opinion a brittle / risky point that can benefit from more real test cases and integration tests. In addition, before this PR, there is already code assumes or uses mates, and it is outside the diff for review. I would encourage double checking the correctness of what you added given the existing code (see the two major issues raised by coderabbitai).

src/constants.py
vratins: I thought it was over-explanatory in this case, since we do mention this later in the code.

src/flow.py
```diff
@@ -716,12 +728,20 @@ def _sample_waters(
         """Dispatch to the configured sampling strategy, sampling one water per
         entry of batch_w and returning them in that order."""
         if self.sampling_strategy == "uniform_ball":
+            # With crystal mates in the graph, anchor on ASU atoms only: the
+            # targets are ASU-only, so mate anchors just disperse the prior.
+            # Runs without mates pass no mask and are unchanged.
+            is_mate = getattr(batch_data["protein"], "is_mate", None)
+            anchor_mask = (
+                ~is_mate.bool() if is_mate is not None and bool(is_mate.any()) else None
+            )
```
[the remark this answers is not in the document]
vratins: Sync removed. New `_asu_mask` returns `~is_mate` with no `.any()`

src/dataset.py
```diff
@@ -914,6 +1058,76 @@ def _parse_pdb_list(self, pdb_list_file: str) -> list[dict]:
...
+            # and a reader must never catch a half-written sidecar.
+            tmp_path = meta_path.with_suffix(f".{os.getpid()}.tmp")
+            with open(tmp_path, "w") as f:
+                json.dump(current, f, indent=2)
+            tmp_path.replace(meta_path)
+        elif any(self.geometry_dir.glob("*.pt")):
+            logger.warning(
+                f"{self.geometry_dir} has no {FILTER_META_FILENAME}; the settings "
+                "its entries were built with cannot be verified."
+            )
```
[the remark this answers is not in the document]
vratins: Addressed

src/dataset.py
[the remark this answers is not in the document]
vratins: Addressed with a comment.

tests/test_dataset.py
```diff
@@ -2665,6 +2889,156 @@ def test_num_asu_protein_metadata_correct(
...
+        )
+        is_mate, emb_res_idx = data["protein"].is_mate, cached["emb_res_idx"]
+        num_asu = data.num_asu_protein_atoms
+
+        assert is_mate.shape == emb_res_idx.shape == (data["protein"].num_nodes,)
+        assert not is_mate[:num_asu].any()
+        assert is_mate.sum().item() > 0
+        # Ligands ride behind the mates, so the mate block is contiguous but need
+        # not run to the end; every marked node is past the ASU.
+        assert is_mate[num_asu:][: is_mate.sum().item()].all()
```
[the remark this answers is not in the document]
vratins: Rewritten to assert the mate-protein block starts at `num_asu` and is contiguous `(diff()==1)`

README.md
vratins: This will likely be better off in another PR where we rename all variables to something more generic, off issue #32

src/dataset.py
vratins: Addressed

src/dataset.py
vratins: Addressed

src/dataset.py
vratins: It's only a warning here, and comes from when pymol parses something biotite dropped, or with an unparsable resi

vratins: We still keep the geometry and the element one hots

src/dataset.py
vratins: Fixed

src/dataset.py
vratins: 1. With the mates model,  the mates contribute to the sampling, so commented where relevant.
2. Understood, deferred for now with a comment

src/flow.py
vratins: Added a warning

src/flow.py
vratins: Added a mask

tests/test_dataset.py
vratins: Addressed.

tests/test_dataset.py
vratins: Done

src/dataset.py
vratins: Added comments for this, but 1) to ensure we drop whole ligands but do not drop whole residues (since we still get some context from partial residues and the esm embeddings for the same) , and 2) to dedup for mates landing on an ASU atom, and then mates landing on another mate (probably a more efficient way to do this whole thing, i agree)

tests/test_dataset.py
```diff
@@ -2952,6 +3119,134 @@ def test_mate_waters_never_enter_the_graph(
...
+        data = self._dataset(tmp_path, pdb_base_dir)[0]
+        cached = torch.load(
+            tmp_path / "processed" / "geometry_mates" / "4h0b_final.pt",
+            weights_only=False,
+        )
+        blocks = self._blocks(data["protein"].is_mate, cached["is_ligand"])
+
+        assert data.num_asu_protein_atoms == int(blocks["asu_protein"].sum())
+        assert int(blocks["asu_ligand"].sum()) > 0
+        assert data.num_asu_protein_atoms < data["protein"].num_nodes
```
DorisMai: the asserts in this test class seems a bit redundant/tautological. for example, here might be better to have `blocks["asu_ligand"].nonzero().min() >= num_asu_protein_atoms)` that more strongly checks no ASU ligand in asu_protein_atoms. Or as your docstring mentioned, you could also check a raise error from slae/esm.

src/dataset.py
```diff
@@ -1402,17 +1414,23 @@ def _preprocess_one(self, entry: dict, cache_path: Path):
                 mate_pos = torch.tensor(mate_coords, dtype=torch.float32) - center
                 mate_x = element_onehot([a.symbol.upper() for a in mate_atoms])
 
-                # compute mate residue indices (group atoms by actual residue)
+                # Group mate atoms by residue. The key omits the symmetry-object id
+                # (atom.model), so two images of one residue share a group. Harmless
```
DorisMai, line 1426: I am worried that while this is okay for mates when not doing pooling, the grouping/collision is actually an issue for deduping ligands, and you might need `.model` as part of your ligand key. See comment for the new `test_real_mate_ligands_are_kept_or_dropped_whole` you add

tests/test_dataset.py
```diff
@@ -452,6 +480,64 @@ def test_segment_separates_entities(self):
...
+        )
+
+        # now make one entity a symmetry image by referencing it against itself
+        groups = {}
+        for i, atom in enumerate(lig_atoms):
+            groups.setdefault(
+                (atom.chain, atom.resi, getattr(atom, "segi", "")), []
+            ).append(i)
+        assert len(groups) > 1, "need more than one entity to show the others survive"
+        target_key, target_idx = next(iter(groups.items()))
```
DorisMai: Currently one your groups (resn HEM) contains 2 ligands that are symmetry related, due to collision under shared chain/resi/segi (the only difference would be `.model` which currently is not keyed). If this group is used for testing here, your appended target (line 522 below) contains only 1 ligand, which has image_frac 0.5 right on the boundary so dedup would NOT drop it. Even if you add a few more atoms to cross the 0.5 fraction, then the whole group (both ligands) would be dropped, which also seems incorrect. Consider make this test stronger by 1) explicitly check a single ligand entity (ideally more than 1 atom) and 2) check dedup when one or more (partial) of this ligand is concatenated as reference to check your image_frac working as intended.

src/flow.py
```diff
@@ -789,14 +827,30 @@ def compute_sigma_per_graph(
...
         empty = torch.bincount(batch_p, minlength=num_graphs) == 0
         if empty.any():
             raise ValueError(
                 f"Cannot compute sigma for graph(s) {empty.nonzero().flatten().tolist()}: "
                 "they have zero protein atoms."
             )
 
+        # Restrict to the masked atoms (ASU-only). Skip the mask, rather than yield
+        # a degenerate sigma, if it would leave a graph with no atoms -- which the
+        # dataset should never produce, so warn if it happens.
```
DorisMai: code seems quite similar to line 112-122. consider a helper function maybe ?

DorisMai (approved):
One more question on ligand dedup. After addressing that I think it's ok to merge.

tests/test_dataset.py
vratins: Done

tests/test_dataset.py
vratins: Reworked the test around this. It now keys entities the way dedup does (.model included), so the symmetry-related HEM pair are two separate entities.

src/flow.py
vratins: Done

src/dataset.py
vratins: `dedup_mate_ligands_by_residue` now keys on `(atom.model, atom.chain, atom.resi, atom.segi)`

PR #89, opened 2026-08-04
Edge type flags

vratins:
- Makes edge construction configurable: `etypes`, `cutoff`, `max_neighbors`, `dynamic_edge_policy`, `knn_fallback_k` on `ProteinWaterUpdate`, which previously hardcoded `ALL_EDGE_TYPES`. This is the dependency `ConfidenceGVP` needs.
 - `get_active_edge_types(disable_ww, disable_wp)` to ablate WW/WP edges; PW and PP always stay on.
 - `build_knn_edges` → `build_dynamic_edges`, supporting both radius and kNN queries, plus a fallback that reconnects nodes a radius query left with no edges in the case of sampling with a scaled gaussian.
 - Replaces `--k_pw`/`--k_ww` with `--dynamic_edge_policy`, `--cutoff`, `--max_neighbors`, `--knn_fallback_k`, `--disable_ww`, `--disable_wp`; `inference.py` reads the same keys back from `config.json`.

## Summary by CodeRabbit

- **New Features**
  - Added configurable dynamic graph edges using radius, KNN, or isolated-node rescue policies.
  - Added controls for distance cutoffs, fallback neighbors, sampling strategy, and per-edge neighbor limits.
  - Added options to independently disable water–water and water–protein connections.
  - Water-related edges can refresh during each forward pass, while protein–protein edges remain cached.

- **Documentation**
  - Updated configuration and training guidance, including legacy setting mappings and new command-line options.

- **Bug Fixes**
  - Improved handling of isolated nodes and invalid edge-policy configurations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,25 +234,25 @@ construction, so training and inference always agree:
 
 | `--dynamic_edge_policy` | Behaviour |
 |-------------------------|-----------|
-| `radius` (default) | Connect every pair within `--cutoff`, capped at `--max_neighbors` per source |
+| `auto` (default) | Resolves off the prior: `radius` under `uniform_ball`, `knn_if_isolated` under `scaled_gaussian` |
+| `radius` | Connect every pair within `--cutoff`, capped at `--max_neighbors` per source |
 | `knn` | Connect a fixed number of nearest neighbours (`--k_pw`, `--k_ww`, `--k_wp`) |
+| `knn_if_isolated` | A `radius` graph plus a KNN rescue for any node the cutoff stranded |
 
-The two differ in which side the neighbour budget applies to. KNN queries *per
-destination*, so every destination is guaranteed edges but a source may have
-none — coverage checks must read the destination row. Radius guarantees nothing:
-a water with no protein atom inside `--cutoff` gets no PW edges at all.
+`radius` and `knn` differ in which side the neighbour budget applies to. KNN
+queries *per destination*, so every destination is guaranteed edges but a source
+may have none — coverage checks must read the destination row. Radius guarantees
+nothing: a water with no protein atom inside `--cutoff` gets no PW edges at all.
 
-`--knn_fallback_k` repairs that. Under `radius`, any water the query stranded is
-reconnected to that many nearest protein atoms regardless of distance. Set it to
-`0` to disable. It has no effect under `knn`, which cannot strand a node.
+`knn_if_isolated` repairs that: any water the radius query stranded is
+reconnected to its `--knn_fallback_k` nearest protein atoms regardless of
+distance (`0` disables the rescue). Plain `radius` does *not* rescue, and the
+flag has no effect under `knn`, which cannot strand a node. `auto` picks
+`knn_if_isolated` for `scaled_gaussian` precisely because Gaussian samples can
+land outside every cutoff, whereas uniform-ball samples cannot.
 
 Set `--disable_ww` / `--disable_wp` to ablate those edge types; PW and PP are
 always active.
-
-> Configs written before the radius/KNN split recorded a three-valued
-> `dynamic_edge_policy` (`auto`, `radius`, `knn_if_isolated`). All three built a
-> radius graph, so they load and map to `radius`; whether stranded waters are
-> rescued is now `--knn_fallback_k`'s job.
 
 ## Embedding Generation
 
@@ -327,9 +327,10 @@ To resume training from a checkpoint, you can load the model weights and optimiz
 | `--scheduler` | `cosine` | LR scheduler: `cosine`, `step`, or `none` |
 | `--warmup_steps` | `0` | Linear warmup steps |
 | `--processed_dir` | `~/flow_cache/` | Cache directory for preprocessed data |
-| `--dynamic_edge_policy` | `radius` | How water-touching edges are built: `radius` or `knn` (see [Edge Construction](#edge-construction)) |
+| `--sampling_strategy` | `uniform_ball` | Flow prior: `uniform_ball` or `scaled_gaussian`; also resolves `--dynamic_edge_policy auto` |
+| `--dynamic_edge_policy` | `auto` | How water-touching edges are built: `auto`, `radius`, `knn`, or `knn_if_isolated` (see [Edge Construction](#edge-construction)) |
 | `--cutoff` | `8.0` | Distance cutoff in Å for radius edges |
-| `--knn_fallback_k` | `8` | Nearest neighbours attached to waters stranded by the radius query; `0` disables |
+| `--knn_fallback_k` | `8` | Nearest neighbours attached to waters stranded by the radius query under `knn_if_isolated`; `0` disables |
 | `--disable_ww` | `false` | Ablate water→water edges |
 | `--disable_wp` | `false` | Ablate water→protein edges |
 | `--include_mates` | `false` | Include symmetry mate atoms as protein nodes |
--- a/README.md
+++ b/README.md
@@ -83,10 +83,13 @@ WaterFlow processes structure files through several stages to create training-re
 - Ligand atoms are appended after ASU and mate atoms and carry the boolean `is_ligand` mask plus `residue_index = -1` (they have no residue embedding, so residue pooling masks them out)
 - `is_mate` marks every non-ASU node, protein or ligand. The flow prior anchors on `~is_mate` so sampled waters start where the targets live
 - Edge types (defined in `src/constants.py`):
-  - `('protein', 'pp', 'protein')`: protein-protein edges
-  - `('protein', 'pw', 'water')`: protein to water
-  - `('water', 'wp', 'protein')`: water to protein
-  - `('water', 'ww', 'water')`: water-water edges
+  - `('protein', 'pp', 'protein')`: protein-protein edges — cached at preprocessing
+  - `('protein', 'pw', 'water')`: protein to water — built at runtime
+  - `('water', 'wp', 'protein')`: water to protein — built at runtime, ablatable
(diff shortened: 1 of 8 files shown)


src/flow.py
```diff
@@ -168,30 +203,42 @@ def sample_waters_scaled_gaussian(
...
+      gets no edges, which is what :meth:`ProteinWaterUpdate._add_knn_fallback`
+      exists to repair.
 
     For a homogeneous graph (``src_pos is dst_pos``) self-edges are dropped.
 
     Args:
         src_pos: (N_src, 3) source node positions.
         dst_pos: (N_dst, 3) destination node positions.
-        k: Number of nearest source neighbors to find per destination node.
+        policy: One of DYNAMIC_EDGE_POLICIES.
```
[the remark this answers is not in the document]
vratins: Addressed

src/flow.py
```diff
@@ -252,12 +331,56 @@ def __init__(
...
+        resolved = resolve_edge_policy(dynamic_edge_policy, sampling_strategy)
+        # Same edges as "radius"; the difference is the extra pass for nodes left with none.
+        self.rescue_isolated = resolved == "knn_if_isolated" and knn_fallback_k > 0
+        self.dynamic_edge_policy = (
+            "radius" if resolved == "knn_if_isolated" else resolved
+        )
+        self.knn_fallback_k = knn_fallback_k
+        self.k_pw = k_pw
+        self.k_ww = k_ww
+        self.k_wp = k_wp
```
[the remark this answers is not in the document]
vratins: Addressed

README.md
```diff
@@ -81,10 +81,13 @@ WaterFlow processes structure files through several stages to create training-re
...
-  - `('protein', 'pw', 'water')`: protein to water
-  - `('water', 'wp', 'protein')`: water to protein
-  - `('water', 'ww', 'water')`: water-water edges
+  - `('protein', 'pp', 'protein')`: protein-protein edges — cached at preprocessing
+  - `('protein', 'pw', 'water')`: protein to water — built at runtime
+  - `('water', 'wp', 'protein')`: water to protein — built at runtime, ablatable
+  - `('water', 'ww', 'water')`: water-water edges — built at runtime, ablatable
+- Only PP edges are stored in the geometry cache; every water-touching edge is
+  rebuilt each forward pass, since water positions move during integration. See
+  [Edge Construction](#edge-construction)
```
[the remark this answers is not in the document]
vratins: The statement is asccurate, every water touching edge is rebuilt every forward pass. PW caching may help only in the case of the confidence model

src/flow.py
```diff
@@ -279,97 +403,177 @@ def __init__(
...
-                pos_w, pos_p, k=k_pw, batch_src=batch_w, batch_dst=batch_p
-            )
-            ei_wp_reversed = ei_wp.flip(0)
-            # union
-            ei_pw_union = torch.cat([ei_pw, ei_wp_reversed], dim=1).unique(dim=1)
-            edge_index_dict[EDGE_PW] = ei_pw_union
-        else:
-            edge_index_dict[EDGE_PW] = torch.empty(
-                2, 0, dtype=torch.long, device=device
+        # Protein-water and water-protein edges get the extra pass; water-water does not.
```
DorisMai: why does ww edge not get the rescue? also I see that the isolate_axis is different for pw vs wp, might worth a comment in `__init__` to explain this asymmetry, as these two fallbacks are really to ensure that all waters always have neighbors to pass or receive messages.

scripts/inference.py
```diff
@@ -273,8 +273,17 @@ def build_model_from_config(config: dict, device: torch.device) -> nn.Module:
         drop_rate=config.get("drop_rate", 0.1),
         n_message_gvps=config.get("n_message_gvps", 2),
         n_update_gvps=config.get("n_update_gvps", 2),
-        k_pw=config.get("k_pw") or 16,
-        k_ww=config.get("k_ww") or 16,
+        cutoff=config.get("cutoff", 8.0),
+        max_neighbors=config.get("max_neighbors", 256),
+        dynamic_edge_policy=config.get("dynamic_edge_policy", "radius"),
```
DorisMai, line 280: presumably at this point you are only dealing with new training runs that used "radius"? otherwise the old config presumably doesn't save this info, and you don't have cli args in the inference script to overwrite this.

src/flow.py
```diff
@@ -279,97 +403,177 @@ def __init__(
...
-        For protein->water edges, we take the union of:
-        - KNN(protein -> water): k nearest waters per protein
-        - KNN(water -> protein) reversed: k nearest proteins per water
-        This ensures every water has at least k_pw protein neighbors.
+    def build_edges(self, data: HeteroData) -> dict[tuple[str, str, str], torch.Tensor]:
+        """
+        Build the edge set for one batch under the active policy.
 
-        PP edges are read from the dataset (cached at preprocessing time).
+        PP and PW edges are read from the dataset when cached at preprocessing
```
DorisMai, line 541: I thought you only cache PP edges because water position changes? Why adding PW cache reading here, do you ever save it as cache?

src/flow.py
```diff
@@ -694,7 +909,6 @@ def __init__(
         self.loss_eps = loss_eps
         self.graph_cutoff = getattr(model, "cutoff", 8.0)
```
DorisMai: Outside the diff, but I traced it here when trying to figure out whether the edge cutoff and the uniform ball cutoff are the same according to the docstring of `resolve_edge_policy`. Then I am wondering if you ever allow setting the "cutoff" attribute of model anywhere? If not, maybe extract the hardcoded 8.0 value out as a constant that explicitly shows the intent (for both uniform ball sampling and radius edge).

src/flow.py
```diff
@@ -448,10 +644,17 @@ def __init__(
         n_message_gvps: int = 2,
         n_update_gvps: int = 2,
         vector_gate: bool = True,
+        water_input_dim: int = 16,  # 1 hot with oxygen, same as encoder
+        cutoff: float = 8.0,
```
DorisMai: it looks like this cutoff value is intended to be reused as `graph_cutoff` when doing uniform ball sampling, which at line 910 (outside diff) also defaults to 8.0 if the cutoff here is not set. Consider extract this as constant to enforce consistency over hard coded value. Also, is this in any way related to the constant `RBF_CUTOFF` which also happens to be 8.0? If the radial edge cutoff > `RBF_CUTOFF`, do you get unintended zeroing of features?

tests/test_flow.py
```diff
@@ -278,20 +365,116 @@ def test_init(self):
...
+
+        def stranded(knn_fallback_k):
+            updater = ProteinWaterUpdate(
+                hidden_dims=(32, 4),
+                layers=1,
+                cutoff=8.0,
+                dynamic_edge_policy="knn_if_isolated",
+                knn_fallback_k=knn_fallback_k,
+            )
+            edge_index = updater.build_edges(data)[EDGE_PW]
```
DorisMai, line 451: consider also have a test for the fallback of WP edge

tests/test_train_config.py
```diff
@@ -131,6 +131,33 @@ def test_inference_build_model_from_config_uses_embedding_dim(device):
     assert model.encoder.output_dims == (128, 0)
 
 
+def test_inference_build_model_from_config_replays_recorded_edge_policy(device):
+    """Every recorded config carries "auto". Replaying one must build a model,
+    not raise, and must land on the radius path those runs actually used."""
+    config = {
```
DorisMai, line 137: do you not need to pass sampling_strategy here and check rescue? if training does scaled_gaussian and "auto" resolves to knn_if_isolated, but inference is radius, is this ok?

tests/test_forward.py
```diff
@@ -183,8 +183,6 @@ def test_forward_pass_no_nan_with_module_hooks(device):
         encoder=encoder,
         hidden_dims=(64, 8),
         layers=2,
-        k_pw=8,  # keep <= n_water_per
-        k_ww=8,  # keep <= n_water_per
```
DorisMai, line 187: there are a few more traces of these k_pw and k_ww arguments in this test file (lines 266, 334, and 379). If you didn't mean to test explicitly on knn as the edge policy, consider clean those up as well.

DorisMai (approved):
A few minor questions/comments, address them if you can before merging.

src/flow.py
vratins: WW edges are mostly vestigial now, since they do not actually work -- I will make sure this is consistent tho. When we were using WW, WP edges, the point was to make sure no water is left without a connection to a protein node, a water not being connected to another water is not an issue, on the other hand.

scripts/inference.py
vratins: Right — new runs record dynamic_edge_policy/sampling_strategy/cutoff in the config. Pre-flag checkpoints fall back to the historical defaults (radius / uniform_ball / 8.0)

src/flow.py
vratins: The PW caching mechanism exists if one ever wanted to use it with the confidence pipeline where candidates are fixed (although impractical at the candidate ratio we use, faster to just construct the edges than the I/O). To be clear, I have never cached the PW for any run -- also very memory heavy.

src/flow.py
vratins: hacky fix right now, but declared a constant default edge distance.

tests/test_flow.py
vratins: Added `test_radius_strands_far_water_on_wp_and_fallback_rescues_it`

tests/test_train_config.py
vratins: Addressed

tests/test_forward.py
vratins: Cleaned them up.

PR #90, opened 2026-08-05
Confidence Model

vratins:
Implementing the confidence model.

- **Targets** — `smootherstep_target`: soft cutoff on nearest-GT distance, 1 inside 0.5 Å, 0 outside 1.5 Å. 
- **Clustering** — `cluster_waters_vdw`: absorb into confidence-weighted centroids at 1.52 Å (oxygen vdW), then NMS over those centroids. Centroids carry the cluster max.
- **`ConfidenceGVP`** — `FlowWaterGVP`'s backbone minus time and self-conditioning, one scalar per candidate, PW+PP edges only. Backbone stays structurally identical so the scorer warm-starts from a flow checkpoint. This needs the dynamic edge type policy and is hence built off PR #89 

## Summary by CodeRabbit

* **New Features**
  * Added confidence scoring for candidate water positions.
  * Added smooth confidence targets and van der Waals–based clustering to filter and rank predictions.
  * Added support for probability or raw confidence-score outputs.
  * Added configurable fallback behavior for model cutoff settings.

* **Documentation**
  * Updated project structure documentation to include the new confidence module and tests.

### DIFF
--- a/src/confidence.py
+++ b/src/confidence.py
@@ -1,0 +1,411 @@
+"""
+Confidence model and post-processing for candidate waters.
+
+Mirrors the SuperWater (Nature Comm. Chem. s42004-025-01789-4) confidence +
+clustering stage that sits after the generator in the DiffDock-style two-stage
+pipeline.
+
+This module provides:
+- smootherstep_target / smootherstep_confidence: per-candidate supervision from
+  a soft cutoff of the nearest-GT-water distance
+- cluster_waters_vdw: vdW-radius clustering with confidence-weighted centroids
+  and NMS over those centroids
+- ConfidenceGVP: scores candidate waters through the same GVP backbone as
+  FlowWaterGVP, minus time conditioning
+"""
+
+from __future__ import annotations
+
+import torch
+import torch.nn.functional as F
+from torch import nn, Tensor
+from torch_geometric.data import HeteroData
+
+from src.constants import EDGE_PP, EDGE_PW, NUM_RBF
+from src.encoder_base import BaseProteinEncoder
+from src.flow import ProteinWaterUpdate
+from src.gvp import GVP
+
+
+# ---------------------------------------------------------------------------
+# Targets
+# ---------------------------------------------------------------------------
+
+
+def _nearest_gt_distance(candidate_pos: Tensor, gt_pos: Tensor) -> Tensor:
+    """
+    Distance from each candidate to its nearest ground-truth water.
+
+    Single-protein (no batch dim); the caller guards the empty cases.
+
+    Args:
+        candidate_pos: (Nc, 3) candidate positions.
+        gt_pos: (Ng, 3) ground-truth positions.
+
+    Returns:
+        (Nc,) Euclidean distances.
+    """
+    diffs = candidate_pos.unsqueeze(1) - gt_pos.unsqueeze(0)  # (Nc, Ng, 3)
+    return diffs.norm(dim=-1).min(dim=1).values
+
+
+def smootherstep(x: Tensor) -> Tensor:
+    """
+    Perlin smootherstep S(x) = 6x^5 - 15x^4 + 10x^3 on x in [0, 1].
+
+    C2-continuous: S, S' and S'' all vanish at both ends, so it joins flat
+    plateaus with no kink in value, slope, or curvature.
+
+    Args:
+        x: Input, already clamped to [0, 1] by the caller.
+
+    Returns:
+        S(x), same shape as `x`.
+    """
+    # Horner form: degree-5 polynomial, no transcendentals.
+    return x * x * x * (10.0 + x * (x * 6.0 - 15.0))
+
+
+def smootherstep_confidence(
+    d: Tensor,
+    r_in: float = 0.5,
+    r_out: float = 1.5,
+) -> Tensor:
+    """
+    Soft-cutoff confidence from a nearest-GT distance.
+
+        conf = 1                    for d <= r_in
+        conf = 1 - smootherstep(u)  for r_in < d < r_out, u = (d-r_in)/(r_out-r_in)
+        conf = 0                    for d >= r_out
+
+    A candidate within `r_in` of a GT water is on the site within experimental
+    error; past `r_out` it is background. The 0.5-crossing sits at the midpoint
+    and the band width sets the steepness, so location and sharpness are
+    decoupled.
+
+    Args:
+        d: (N,) nearest-GT distances in Angstroms.
+        r_in: Plateau radius -- at or under this, confidence is 1.
+        r_out: Floor radius -- at or over this, confidence is 0.
+
+    Returns:
+        (N,) confidences in [0, 1].
+
+    Raises:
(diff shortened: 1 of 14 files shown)


src/confidence.py
```diff
@@ -0,0 +1,411 @@
...
+        """
+        device = data["protein"].pos.device
+
+        s_all, v_all, pp_edge_attr = self.encoder(data)
+        encoder_input = (s_all, v_all) if self.encoder.output_dims[1] > 0 else s_all
+        s_p_latent, v_p_latent = self.encoder_to_flow(encoder_input)
+
+        if "water" not in data.node_types or data["water"].num_nodes == 0:
+            return torch.zeros(0, device=device)
+
```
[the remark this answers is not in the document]
vratins: Moved the guard above the encoder call

src/confidence.py
```diff
@@ -0,0 +1,411 @@
...
+        d2 = ((pos_sorted - seed_pos) ** 2).sum(dim=-1)
+        # Always includes i: d2[i] == 0 and assigned[i] is False.
+        within = (d2 <= r2) & (~assigned)
+        cluster_pos = pos_sorted[within]
+        cluster_conf = conf_sorted[within]
+
+        # Fall back to an unweighted mean when the weights sum to 0, so an
+        # all-zero-confidence cluster yields a position rather than NaN.
+        w_sum = cluster_conf.sum()
+        if w_sum.abs() > 0:
```
DorisMai: why abs()? w_sum > 0 should be sufficient? do you expect to have negative confidence? if so, .abs() still seems a wrong fix.

src/confidence.py
```diff
@@ -0,0 +1,411 @@
...
+        d2 = ((cent_pos - cent_pos[idx]) ** 2).sum(dim=-1)
+        collide = (d2 <= r2) & keep_mask
+        collide[idx] = False  # never drop self
+        keep_mask[collide] = False
+
+    cent_pos = cent_pos[keep_mask]
+    cent_conf = cent_conf[keep_mask]
+
+    # Descending order gives downstream thresholding a stable prefix.
+    final_order = torch.argsort(cent_conf, descending=True)
```
DorisMai: seems unnecessary here and line 228 given you already sorted in line 186? can probably just add stable=True there if you can exact ordering on ties

tests/test_confidence.py
```diff
@@ -0,0 +1,420 @@
...
+
+    def test_two_far_waters_preserved(self, device):
+        # 2 A apart with r=1.52 => kept separate.
+        pos = torch.tensor([[0.0, 0.0, 0.0], [2.0, 0.0, 0.0]], device=device)
+        conf = torch.tensor([0.9, 0.8], device=device)
+        out_pos, out_conf = cluster_waters_vdw(pos, conf, radius=1.52)
+        assert out_pos.size(0) == 2
+        assert out_conf[0].item() >= out_conf[1].item()
+
+    def test_threshold_filters_pre_cluster(self, device):
```
DorisMai, line 120: seems redundant to the test below. if you wanted to assert on the confidence, can probably add to the test below or test_cluster_confidence_is_the_max_not_the_mean. When adding, consider have one water exactly on the threshold (e.g. confidence == 0.5)

src/confidence.py
```diff
@@ -0,0 +1,411 @@
...
+    (inputs are clean samples), emits one scalar per candidate water, and runs
+    on PW and PP edges only -- candidates come from inference and are not
+    refined here, so WW and WP carry nothing.
+
+    PW edges come from the dataset when it supplies them, otherwise from a
+    radius query plus a nearest-neighbour pass for candidates it left with no
+    edges -- candidates can land in sparse regions the generator never visits.
+
+    Keeping the backbone identical to the generator is what lets the confidence
+    network warm-start from a flow checkpoint.
```
DorisMai: make a note/comment maybe in a future pr on how shape mismatch for warm-start is addressed since the protein scalar encoder no longer has time conditioning?

src/confidence.py
```diff
@@ -0,0 +1,411 @@
...
+            drop_rate=drop_rate,
+            n_message_gvps=n_message_gvps,
+            n_update_gvps=n_update_gvps,
+            vector_gate=vector_gate,
+            aggr_edges="sum",
+            use_dst_feats=True,
+            etypes=[EDGE_PW, EDGE_PP],
+            cutoff=cutoff,
+            max_neighbors=max_neighbors,
+            dynamic_edge_policy=dynamic_edge_policy,
```
DorisMai, line 358: should reject "auto" dynamic_edge_policy as input before reaching here. Otherwise it will resolve to "radius". Also does it work for other policy like "knn", since parameters like k_pw are not plumbed through?

tests/test_confidence.py
```diff
@@ -0,0 +1,420 @@
...
+        assert out_pos[0, 0].item() == pytest.approx(expected_x, abs=1e-5)
+        assert out_conf[0].item() == pytest.approx(0.9, abs=1e-6)  # cluster max
+
+    def test_two_far_waters_preserved(self, device):
+        # 2 A apart with r=1.52 => kept separate.
+        pos = torch.tensor([[0.0, 0.0, 0.0], [2.0, 0.0, 0.0]], device=device)
+        conf = torch.tensor([0.9, 0.8], device=device)
+        out_pos, out_conf = cluster_waters_vdw(pos, conf, radius=1.52)
+        assert out_pos.size(0) == 2
+        assert out_conf[0].item() >= out_conf[1].item()
```
DorisMai: should assert on pos (cluster pos are seed pos in this case)

tests/test_confidence.py
```diff
@@ -0,0 +1,420 @@
...
+        scores = model(data)
+        target = torch.rand_like(scores)
+        loss = F.mse_loss(scores, target)
+        loss.backward()
+
+        has_grad = any(
+            p.grad is not None and p.grad.abs().sum().item() > 0
+            for p in model.parameters()
+        )
+        assert has_grad
```
DorisMai: seems trivial given the score_head grad check in the test below. maybe explicitly look at the gradients of a warm-started part of the model like encoder?

DorisMai (approved):
mostly minor comments, should be easy to address.

src/confidence.py
vratins: good catch, addressed.

src/confidence.py
vratins: Yep, done.

tests/test_confidence.py
vratins: Done

src/confidence.py
vratins: clarified in the docstring.

src/confidence.py
vratins: Added a guard here

tests/test_confidence.py
vratins: Done

tests/test_confidence.py
vratins: Yep, now checks a warm-started part (the encoder) receives gradient

PR #94, opened 2026-08-13
Adding DDP to the repo for multi-GPU training

vratins:
Multi-GPU flow matching training via torchrun

New `src/distributed.py` (shared DDP helpers) + DDP wiring in `train.py`. No CLI flag — `torchrun` env vars are the only switch, single-GPU runs unchanged. Verified on 2×H100. tests are CPU-only for CI for now.

Updated README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,34 +280,24 @@ uv run python -m scripts.train \
 
 ### Multi-GPU Training (DDP)
 
-Launch with `torchrun`. There is no DDP flag — the launcher's `WORLD_SIZE` /
-`RANK` / `LOCAL_RANK` are the only switch, so a run's `config.json` is identical
-whether it used one GPU or eight.
+No DDP flag — `torchrun`'s env vars are the only switch; a plain
+`python -m scripts.train` runs single-GPU as before.
 
 ```bash
 uv run torchrun --nproc_per_node=4 -m scripts.train \
     --train_list splits/train_list_0.95.txt \
     --val_list splits/valid_list_0.05.txt \
     --encoder_type gvp \
-    --batch_size 4
+    --batch_size 4  # per rank -> effective 16
 ```
 
-`--batch_size` is per rank, so the above trains at an effective batch of 16.
-
-| | Behavior across ranks |
-|---|---|
-| Data | `DistributedSampler` gives each rank a disjoint shard, reshuffled per epoch. The last shard is padded, so a few samples are counted twice in epoch metrics |
-| Gradients | All-reduced once per optimizer step, not per accumulation micro-step |
-| Metrics | Train/val/eval sums are all-reduced, so every rank sees identical averages and agrees on the best epoch |
-| Sampling eval | The fixed eval set is strided across ranks and integrated on the unwrapped model |
-| Disk and W&B | Rank 0 only: `config.json`, checkpoints, `train.log`, plots, W&B |
-
-The geometry cache is built before the process group exists, by rank 0 alone,
-with the other ranks blocked on a CPU-side store — so a cold build (which can
-take hours) never counts against an NCCL timeout. A warm cache makes it a no-op.
-
-> Checkpoints hold the unwrapped `state_dict`, so `inference.py` loads a
-> DDP-trained checkpoint with no key rewriting.
+- Each rank trains on a disjoint `DistributedSampler` shard, reshuffled per epoch.
+- Gradients all-reduce once per optimizer step; train/val/eval metrics are
+  all-reduced, so every rank agrees on the best epoch.
+- Rank 0 owns disk and W&B (config, checkpoints, logs). Checkpoints hold the
+  unwrapped `state_dict`, so `inference.py` loads them unchanged.
+- The geometry cache is built by rank 0 before the NCCL group exists,
+  coordinated on a CPU-side store — a cold build can't trip a collective timeout.
 
 ### Resuming from Checkpoints
 
--- a/README.md
+++ b/README.md
@@ -306,8 +306,10 @@ uv run python -m scripts.train \
 
 ### Multi-GPU Training (DDP)
 
-No DDP flag — `torchrun`'s env vars are the only switch; a plain
-`python -m scripts.train` runs single-GPU as before.
+To train on several GPUs on one machine, launch the same script with `torchrun`
+and set `--nproc_per_node` to the number of GPUs you want to use. Launching with 
+`torchrun` turns on multi-GPU training, and the plain `python -m scripts.train` 
+command still trains on a single GPU.
 
 ```bash
 uv run torchrun --nproc_per_node=4 -m scripts.train \
@@ -316,14 +318,6 @@ uv run torchrun --nproc_per_node=4 -m scripts.train \
     --encoder_type gvp \
     --batch_size 4  # per rank -> effective 16
 ```
-
-- Each rank trains on a disjoint `DistributedSampler` shard, reshuffled per epoch.
-- Gradients all-reduce once per optimizer step; train/val/eval metrics are
-  all-reduced, so every rank agrees on the best epoch.
-- Rank 0 owns disk and W&B (config, checkpoints, logs). Checkpoints hold the
-  unwrapped `state_dict`, so `inference.py` loads them unchanged.
-- The geometry cache is built by rank 0 before the NCCL group exists,
-  coordinated on a CPU-side store — a cold build can't trip a collective timeout.
 
 ### Resuming from Checkpoints
 
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ WaterFlow/
 │   ├── gvp.py              # Geometric Vector Perceptron layers
 │   ├── gvp_encoder.py      # GVP-based protein encoder
 │   ├── encoder_base.py     # Encoder registry and factory (includes ESM/SLAE)
+│   ├── distributed.py      # DDP helpers (rank discovery, collectives, barriers)
 │   ├── constants.py        # Shared constants (RBF bins, etc.)
 │   └── utils.py            # Metrics, plotting, logging utilities
 ├── scripts/                # Executable scripts
@@ -21,6 +22,7 @@ WaterFlow/
 │   └── generate_slae_embeddings.py  # Precompute SLAE embeddings
 ├── tests/                  # Test suite
 │   ├── test_dataset.py     # Dataset and preprocessing tests
+│   ├── test_distributed.py # DDP helper and cache prebuild tests
 │   ├── test_flow.py        # Flow matching tests
 │   ├── test_encoder.py     # Encoder tests
 │   ├── test_forward.py     # End-to-end forward pass tests
@@ -275,6 +277,37 @@ uv run python -m scripts.train \
     --grad_accum_steps 4 \
     --processed_dir ~/flow_cache/
 ```
+
+### Multi-GPU Training (DDP)
+
+Launch with `torchrun`. There is no DDP flag — the launcher's `WORLD_SIZE` /
+`RANK` / `LOCAL_RANK` are the only switch, so a run's `config.json` is identical
+whether it used one GPU or eight.
+
+```bash
+uv run torchrun --nproc_per_node=4 -m scripts.train \
+    --train_list splits/train_list_0.95.txt \
+    --val_list splits/valid_list_0.05.txt \
+    --encoder_type gvp \
+    --batch_size 4
+```
+
+`--batch_size` is per rank, so the above trains at an effective batch of 16.
+
+| | Behavior across ranks |
+|---|---|
+| Data | `DistributedSampler` gives each rank a disjoint shard, reshuffled per epoch. The last shard is padded, so a few samples are counted twice in epoch metrics |
+| Gradients | All-reduced once per optimizer step, not per accumulation micro-step |
+| Metrics | Train/val/eval sums are all-reduced, so every rank sees identical averages and agrees on the best epoch |
+| Sampling eval | The fixed eval set is strided across ranks and integrated on the unwrapped model |
+| Disk and W&B | Rank 0 only: `config.json`, checkpoints, `train.log`, plots, W&B |
+
+The geometry cache is built before the process group exists, by rank 0 alone,
+with the other ranks blocked on a CPU-side store — so a cold build (which can
+take hours) never counts against an NCCL timeout. A warm cache makes it a no-op.
+
+> Checkpoints hold the unwrapped `state_dict`, so `inference.py` loads a
+> DDP-trained checkpoint with no key rewriting.
 
 ### Resuming from Checkpoints
 
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -771,16 +771,6 @@ def run_eval_sampling(
     return avg_metrics
 
 
-def _collective_device(args: argparse.Namespace) -> torch.device:
-    """
-    Device for DDP collective buffers.
-
-    NCCL requires this rank's own CUDA device; the CPU fallback only ever runs
-    single-process, where the collective is a no-op.
-    """
-    return torch.device(args.device if torch.cuda.is_available() else "cpu")
-
-
 def _needs_grad_sync(step: int, n_batches: int, accum_steps: int) -> bool:
     """
     Whether this micro-step's backward must all-reduce gradients under DDP.
@@ -800,6 +790,7 @@ def train_epoch(
     optimizer: AdamW,
     warmup_scheduler,
     args: argparse.Namespace,
+    device: torch.device,
     epoch: int,
     optimizer_step_count: int,
 ) -> tuple[dict[str, float], int, int]:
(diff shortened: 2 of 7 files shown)


scripts/train.py
```diff
@@ -763,11 +811,17 @@ def train_epoch(
             skipped_batches += 1
             continue
```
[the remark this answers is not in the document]
vratins: By construction the data will not have any batches with 0 water and that will be caught early, and this is more of a debug

src/distributed.py
```diff
@@ -0,0 +1,165 @@
...
+
+def is_main_process(rank: int) -> bool:
+    """True on the single rank that owns disk IO and W&B logging."""
+    return rank == 0
+
+
+def ddp_barrier() -> None:
+    """Block until every rank arrives. No-op when not distributed."""
+    if ddp_is_active():
+        dist.barrier()
```
marcuscollins, line 83: I thought dist.barrier did this check already? Or rather, I thought it didn't matter (i.e., DDP still works with a world size == 1)

README.md
```diff
@@ -300,6 +302,27 @@ uv run python -m scripts.train \
     --processed_dir ~/flow_cache/
 ```
 
+### Multi-GPU Training (DDP)
+
+No DDP flag — `torchrun`'s env vars are the only switch; a plain
```
marcuscollins: This is cryptic AI-speak. If someone doesn't know what `torchrun` is, this doesn't help them. Rewrite declaratively, just tell people what to do if they want to use more than one GPU.

README.md
```diff
@@ -300,6 +302,27 @@ uv run python -m scripts.train \
...
+
+```bash
+uv run torchrun --nproc_per_node=4 -m scripts.train \
+    --train_list splits/train_list_0.95.txt \
+    --val_list splits/valid_list_0.05.txt \
+    --encoder_type gvp \
+    --batch_size 4  # per rank -> effective 16
+```
+
+- Each rank trains on a disjoint `DistributedSampler` shard, reshuffled per epoch.
```
marcuscollins: Remove this line.

scripts/train.py
```diff
@@ -1085,37 +1241,56 @@ def main():
...
             num_graphs = int(batch["protein"].batch.max().item()) + 1
             t = torch.zeros(num_graphs, device=device)
-            v_out = model(batch, t)
+            v_out = raw_model(batch, t)
             logger.info(f"Forward pass successful! Output shape: {v_out.shape}")
             logger.info(f"Output stats: mean={v_out.mean():.4f}, std={v_out.std():.4f}")
             if v_out.std() < 1e-6:
                 logger.warning("Model output is constant! This indicates a problem.")
-        model.train()
+        raw_model.train()
```
marcuscollins, line 1293: I think I would still set this on `model`, not just `raw_model`. I don't remember well enough whether there are other things in the DDP object that might need setting.

scripts/train.py
```diff
@@ -1085,37 +1241,56 @@ def main():
...
 
     flow_matcher = FlowMatcher(
         model=model,
```
marcuscollins, line 1296: Does the FlowMatcher need/have updates to understand the DDP object?

scripts/train.py
```diff
@@ -733,17 +756,42 @@ def run_eval_sampling(
...
-            "eval/avg_precision": np.mean([r["precision"] for r in results]),
-            "eval/avg_recall": np.mean([r["recall"] for r in results]),
-            "eval/avg_f1": np.mean([r["f1"] for r in results]),
-            "eval/avg_auc_pr": np.mean([r["auc_pr"] for r in results]),
-        }
-        wandb.log(avg_metrics, step=global_step)
-        return avg_metrics
-    return {}
+    # Every rank must reach this, even one whose stride was all zero-water graphs.
+    avg_metrics, _ = all_reduce_means(
```
marcuscollins, line 760: There is an AVG reduce operation in torch. Why not use it directly with `dist.reduce`?

scripts/train.py
```diff
@@ -770,11 +818,17 @@ def train_epoch(
...
-            batch,
-            use_self_conditioning=args.use_self_cond,
-            accumulation_steps=args.grad_accum_steps,
+        # Suppress the gradient all-reduce on micro-steps that are not followed by
+        # an optimizer.step(), keeping comms at one all-reduce per optimizer step.
+        no_sync = ddp_is_active() and not _needs_grad_sync(
+            step, len(train_loader), args.grad_accum_steps
         )
+        with flow_matcher.model.no_sync() if no_sync else contextlib.nullcontext():
+            metrics = flow_matcher.training_step(
```
marcuscollins, line 818: This feels weird to me... don't you have to set a barrier somewhere and do some reduce operation? It's been a while since I've looked at the low-level DDP package, is the DDP object itself handling the synchronization and reduction of gradients?

src/distributed.py
```diff
@@ -0,0 +1,165 @@
...
+    keys = list(sums)
+    totals = torch.tensor(
+        [float(sums[k]) for k in keys] + [float(count)],
+        dtype=torch.float64,
+        device=device,
+    )
+    if ddp_is_active():
+        dist.all_reduce(totals, op=dist.ReduceOp.SUM)
+
+    total_count = int(totals[-1].item())
```
marcuscollins: Use `total_count = totals[-1].item()` here, so you can re-use it as the denominator below, and then return `int(total_count)` in the return line.

src/distributed.py
```diff
@@ -0,0 +1,165 @@
...
+    if not ddp_is_active():
+        work()
+        return None
+
+    rank = int(os.environ["RANK"])
+    world = int(os.environ["WORLD_SIZE"])
+    host = os.environ["MASTER_ADDR"]
+    port = int(os.environ["MASTER_PORT"])
+    # Client of torchrun's agent store (is_master=False) -- the agent already owns
+    # MASTER_PORT. The timeout covers a cold build, which can take hours.
```
marcuscollins: I don't understand what you're building that could take up to 24 hours? If it is the data set, you shouldn't be doing that every time you run anyway. Anyway it seems like this kind of thing is better handled in your training script, protecting it in a conditional that runs only on rank 0, and using a barrier to prevent the other processes from proceeding until all ranks (in this case only 0) are done.

scripts/train.py
```diff
@@ -847,6 +901,14 @@ def train_epoch(
 
     final_global_step = (epoch - 1) * len(train_loader) + len(train_loader) - 1
 
+    # One collective per epoch so metrics cover every rank's shard. Must run before
```
marcuscollins: Do you mean "one collection"? I don't understand this comment at all.

scripts/train.py
```diff
@@ -847,6 +901,14 @@ def train_epoch(
...
+    # the zero-batch check, or a rank that skipped everything would never enter it.
+    train_metrics, processed_batches = all_reduce_means(
+        {"train/epoch_loss": total_loss, "train/epoch_rmsd": total_rmsd},
+        processed_batches,
+        _collective_device(args),
```
marcuscollins: So, if I understand correctly, _collective_device should return the current rank device this particular instance of the script is running with. This feels clunky and error prone. Usually you want to set the device for your process, and then use that, so that the device is the same as local_rank. I thought torchrun did this by default by setting CUDA_VISIBLE_DEVICES to local_rank when it spawns the subprocesses, so that all you'd need to do here is pass local_rank.

src/distributed.py
```diff
@@ -0,0 +1,165 @@
...
+
+    Returns:
+        The store, for `setup_distributed(store=...)`; None when not distributed
+        (in which case `work` simply runs inline).
+    """
+    if not ddp_is_active():
+        work()
+        return None
+
+    rank = int(os.environ["RANK"])
```
marcuscollins, line 114: why not use `dist.get_rank()` etc. here?

scripts/train.py
```diff
@@ -992,27 +1051,121 @@ def build_scheduler(optimizer, args):
...
+        shutil.rmtree(tmpdir, ignore_errors=True)
+
+
 def main():
     """Run the full training pipeline."""
     args = parse_args()
+
+    # Build the cache before the NCCL group exists: a cold build can take hours and
+    # would trip a GPU-collective timeout. The store is reused as NCCL's rendezvous.
+    store = run_once_on_main(lambda: build_cache(args), key="wf_cache_ready")
```
marcuscollins, line 1130: If this can take hours, and is the same for many training runs, you should not be running it as part of this script at all. Instead, build it once if that is possible, then load the data. Then you can avoid needing the store and any associated timeouts.

scripts/train.py
```diff
@@ -992,27 +1051,121 @@ def build_scheduler(optimizer, args):
...
+    hence race-free. A warm cache is a fast no-op; a cold build is parallelized
+    across CPU cores over disjoint key shards.
+    """
+    dataset_kwargs, _, _ = _build_dataset_config(args)
+    ids = set()
+    for lst in (args.train_list, args.val_list):
+        with open(lst) as f:
+            ids.update(line.strip() for line in f if line.strip())
+    sorted_ids = sorted(ids)
+    if not sorted_ids:
```
marcuscollins: always bail out as soon as you can. You can check if `ids` is empty before sorting.

scripts/train.py
```diff
@@ -992,27 +1051,121 @@ def build_scheduler(optimizer, args):
...
+    Build the geometry cache for the train+val lists as the sole writer.
+
+    Preprocessing is CPU/PyMOL only, so this runs before the DDP group exists --
+    hence race-free. A warm cache is a fast no-op; a cold build is parallelized
+    across CPU cores over disjoint key shards.
+    """
+    dataset_kwargs, _, _ = _build_dataset_config(args)
+    ids = set()
+    for lst in (args.train_list, args.val_list):
+        with open(lst) as f:
```
marcuscollins, line 1075: You can also use `for line in open(list):`, the open method provides the correct context that will close the file when the iterator is exhausted.

scripts/train.py
```diff
@@ -992,27 +1051,121 @@ def build_scheduler(optimizer, args):
...
+            ids.update(line.strip() for line in f if line.strip())
+    sorted_ids = sorted(ids)
+    if not sorted_ids:
+        return
+
+    tmpdir = Path(tempfile.mkdtemp(prefix="wf_build_"))
+    try:
```
marcuscollins, line 1082: There's no except block? I feel like you should at least log the error.

scripts/train.py
```diff
@@ -992,27 +1051,121 @@ def build_scheduler(optimizer, args):
...
+        union.write_text("\n".join(sorted_ids) + "\n")
+        # Parse-only probe to find which entries still need building.
+        probe = ProteinWaterDataset(
+            pdb_list_file=str(union),
+            processed_dir=args.processed_dir,
+            preprocess=False,
+            **dataset_kwargs,
+        )
+        missing = [
+            entry["cache_key"]
```
marcuscollins: There are fewer cache keys than entries, right? If so, make a set of the keys and only check for each file once, rather than for every entry.

scripts/train.py
```diff
@@ -992,27 +1051,121 @@ def build_scheduler(optimizer, args):
...
+            if not (probe.geometry_dir / f"{entry['cache_key']}.pt").is_file()
+        ]
+        if not missing:
+            return  # warm cache: nothing to build
+
+        logger.info(f"build_cache: preprocessing {len(missing)} missing entries")
+        n_shards = max(1, min(len(missing), os.cpu_count() or 1))
+        if n_shards == 1:
+            _build_cache_shard(str(union), args.processed_dir, dataset_kwargs)
+            return
```
marcuscollins: This seems unnecessary. I don't think there's that much overhead if you spawn another process for this task, even if there's just one shard.

scripts/train.py
```diff
@@ -992,27 +1051,121 @@ def build_scheduler(optimizer, args):
...
+        shard_files = []
+        for i in range(n_shards):
+            shard = tmpdir / f"shard_{i}.txt"
+            shard.write_text("\n".join(missing[i::n_shards]) + "\n")
+            shard_files.append(str(shard))
+        # spawn (not fork): safe alongside PyMOL's C extension and any threads.
+        ctx = mp.get_context("spawn")
```
marcuscollins, line 1108: In the future I encourage you to use higher-level interfaces like `joblib` which makes this kind of thing trivially easy. Here it isn't too complex, since _build_cache_shard only raises an error or returns None, but if you care about the output `joblib` makes life much simpler.

scripts/train.py
```diff
@@ -992,27 +1051,121 @@ def build_scheduler(optimizer, args):
...
+
+    # Build the cache before the NCCL group exists: a cold build can take hours and
+    # would trip a GPU-collective timeout. The store is reused as NCCL's rendezvous.
+    store = run_once_on_main(lambda: build_cache(args), key="wf_cache_ready")
+
+    # Under torchrun each rank binds its own GPU; a plain launch yields (0, 0, 1).
+    rank, local_rank, world_size = setup_distributed(store=store)
+    main_proc = is_main_process(rank)
+    if ddp_is_active():
+        args.device = f"cuda:{local_rank}"
```
marcuscollins: Okay, regarding an above comment, don't add this complexity. Just use `local_rank` when you need to figure out what device to send a tensor to. You don't need to append it to `args` for this.

marcuscollins (approved):
The code looks technically correct to me, but there are a lot of things I would change or look into. Given time constraints I'm approving, but please look over the comments and update what you can.

src/distributed.py
vratins: For the barrier to work you need a process group, and dist.barrier() would crash with a single GPU run

README.md
vratins: Done

README.md
vratins: Done

scripts/train.py
vratins: I checked, and nothing else needs setting wit DDP, and setting this on model made the block look inconsistent, hence went with raw_model since that is where the eval flags are being set anyway

scripts/train.py
vratins: FlowMatcher is handed the wrapped model, and custom attributes are handled in flow.py, checked this.

scripts/train.py
vratins: Yes, DDP does that itself, it attaches a hook to every parameter, and during backward(), DDP handles the reduce. No need for manual barrier/reduce for gradients.

scripts/train.py
vratins: AVG uses the mean, but not weighted by the number of samples on a GPU. Technically the sum and then divide we are doing is equivalent to all our training cases since this is an issue only when we have imbalanced batch sizes across different GPUs, and that would be when we skip a batch due to no waters -- but this is taken care of in pre-processing, before training. But I kept sum-then-divide because it stays correct if counts ever diverge as a guard.

src/distributed.py
vratins: Yep, done.

scripts/train.py
vratins: Meant the all-reduce operation. Re-worded.

scripts/train.py
vratins: Done.

src/distributed.py
vratins: `run_once_on_main` runs before `setup_distributed/init_process_group`, so there's no process group yet and `dist.get_rank()` would raise an error. Might change this while addressing the comments about building the dataset before any of this

src/distributed.py
vratins: A 24 hour run would only be if we are running it from scratch, but the code also goes through any missing entries from given split -- clarified comments towards that.

scripts/train.py
vratins: torchrun doesn't set CUDA_VISIBLE_DEVICES, it only sets LOCAL_RANK, and we call torch.cuda.set_device(local_rank) ourselves, so each process sees all GPUs and must target cuda:local_rank (which args.device holds). But yes, the helper is redundant, so I unified device handling (addressing line 1140 too)

scripts/train.py
vratins: Clarified that comment; I think we need the store to safeguard missing caches

scripts/train.py
vratins: I chose to keep it this way since I was getting ruff flags (SIM115)

scripts/train.py
vratins: Added one.

scripts/train.py
vratins: Done

scripts/train.py
vratins: Addressed

scripts/train.py
vratins: `_build_cache_shard` returns None/raises, so there's no output to collect hence went with mp.Pool. Will explore joblib in the future!

scripts/train.py
vratins: Yep, addressed. main() computes once device that's threaded through.

Issue #92, opened 2026-08-13
Remove self-conditioning and late stage time conditioning from repo to clean model load


vratins:
`FlowWaterGVP` expects sc keys in the checkpoint -- none of the latest models are trained with it.

PR #97, opened 2026-08-18
Script and tests to generate candidates for training confidence model

vratins:
- Adding `scripts/cache_candidate.py` to generate the candidate cache that the confidence model trains on, after the flow model has been trained. Writes one cache file per structure with the candidates' coordinates.
- Contains no self-conditioning code, so it should merge after the SC-removal PR is merged (but free to review right now). 
- touches some files where I rename the word "sidecar" to just files for caches for clarity in comments. 

## Summary by CodeRabbit

- **New Features**
  - Added a utility for generating confidence-model candidate caches from trained flow checkpoints.
  - Supports configurable sampling, datasets, devices, checkpoints, output locations, and overwrite behavior.
  - Records generation metadata and statistics for each cache.

- **Documentation**
  - Documented the new cache-generation utility and clarified filter metadata file terminology.

- **Tests**
  - Added coverage for cache creation, metadata, overwrite handling, and run-specific output directories.

### DIFF
--- a/scripts/cache_candidates.py
+++ b/scripts/cache_candidates.py
@@ -109,7 +109,12 @@ def _write_candidate_cache(
 
     (out / "generation.json").write_text(
         json.dumps(
-            {**run_info, "n_written": n_written, "n_skipped": n_skipped, "n_total": n_total},
+            {
+                **run_info,
+                "n_written": n_written,
+                "n_skipped": n_skipped,
+                "n_total": n_total,
+            },
             indent=2,
         )
     )
--- a/scripts/cache_candidates.py
+++ b/scripts/cache_candidates.py
@@ -1,0 +1,353 @@
+"""
+Generate the candidate cache for confidence-model training.
+
+Samples candidate waters from a trained flow checkpoint over the flow dataset
+cache layout and writes one `<pdb_id>.pt = {"candidate_pos": (Nc, 3)}` per
+structure, plus a `generation.json` record of how it was made. Train the confidence
+model on the result with `scripts/train_confidence.py --candidate_dir <out_dir>`.
+
+Everything the confidence model needs (protein graph, embeddings, GT waters, PP edges) 
+is loaded from the flow caches at train time. Model loading and integration reuse the flow inference
+machinery verbatim, so candidates are sampled exactly as `scripts/inference.py`
+would sample them.
+
+Example:
+    python -m scripts.cache_candidates \\
+        --flow_run_dir <run_dir> \\
+        --pdb_list splits/conf_train.txt \\
+        --processed_dir <cache_root> \\
+        --base_pdb_dir <pdb_dir> \\
+        --water_ratio 3.0 --seed 0 --method euler --num_steps 100
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+from pathlib import Path
+
+import torch
+from loguru import logger
+from tqdm import tqdm
+
+from scripts.inference import (
+    _extract_dataset_filter_config,
+    build_model_from_config,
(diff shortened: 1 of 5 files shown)


scripts/cache_candidates.py
```diff
@@ -0,0 +1,358 @@
...
+
+    def sample_batch(graphs):
+        return run_inference_batch(
+            flow_matcher,
+            graphs,
+            method=method,
+            num_steps=num_steps,
+            device=str(device_t),
+            water_ratio=water_ratio,
+        )
```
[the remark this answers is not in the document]
vratins: use_sc is removed in PR #92 and this PR is built off it, hence not passing it here.

scripts/cache_candidates.py
```diff
@@ -0,0 +1,358 @@
...
+            for i in chunk
+            if overwrite or not (out / f"{entries[i]['cache_key']}.pt").exists()
+        ]
+        n_skipped += len(chunk) - len(todo)
+        if not todo:
+            continue
+        for result in sample_batch([dataset[i] for i in todo]):
+            candidate_pos = torch.as_tensor(result["water_pred"], dtype=torch.float32)
+            torch.save({"candidate_pos": candidate_pos}, out / f"{result['pdb_id']}.pt")
+            n_written += 1
```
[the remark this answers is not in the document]
vratins: water_pred in this case (or at this stage) is detached and on the CPU. not a concern. (`x.detach().cpu()[mask].numpy()` in the integrate methods)

scripts/cache_candidates.py
```diff
@@ -0,0 +1,358 @@
...
+        pdb_list: Text file of `<pdb_id>_final` keys, one per line.
+        processed_dir: Cache root shared with flow training (geometry + esm).
+        base_pdb_dir: Base PDB dir, as used by flow training.
+        out_dir: Output dir. Defaults to `default_candidate_dir(...)`.
+        checkpoint: Checkpoint filename under `{flow_run_dir}/checkpoints`.
+        water_ratio: Oversampling ratio — sample `num_residues * water_ratio` waters.
+        seed: RNG seed for the sampling prior (reproducible candidates).
+        num_steps, method: Integration settings, as in inference.
+        batch_size: Graphs per integration batch.
+        geometry_cache_name / include_mates: Optional overrides; default to the
```
DorisMai: technically if flow config has include_mates=True, you currently don't have a way to override it via cli args here

scripts/cache_candidates.py
```diff
@@ -0,0 +1,358 @@
...
+from src.utils import setup_logging_for_tqdm
+
+
+def default_candidate_dir(
+    processed_dir: str | Path,
+    flow_run_dir: str | Path,
+    water_ratio: float,
+    seed: int,
+) -> Path:
+    """Namespaced candidate dir so caches from different checkpoints, ratios, or
```
DorisMai: different checkpoints and integration steps are not reflected in the default dir name though, only saved in generation.json

scripts/cache_candidates.py
```diff
@@ -0,0 +1,358 @@
...
+    )
+
+    out = (
+        Path(out_dir)
+        if out_dir is not None
+        else default_candidate_dir(processed_dir, run_dir, water_ratio, seed)
+    )
+    torch.manual_seed(seed)  # reproducible sampling prior
+
+    def sample_batch(graphs):
```
DorisMai, line 238: consider add a comment above explain why you need a wrapper function here, and spell out the type of `graphs` and the type of output.

DorisMai (approved):

PR #98, opened 2026-08-18
Confidence model training script

vratins:
- Code to train `ConfidenceGVP` to score candidate waters using BCE-with-logits on the smootherstep target function. 
- Validation methods for AUC-PR and best F1
- DDP integrated


scripts/train_confidence.py
```diff
@@ -0,0 +1,815 @@
...
+            "hidden_v": config.get("hidden_v") or 64,
+            "node_scalar_in": config.get("node_scalar_in") or 16,
+            "freeze_encoder": config.get("freeze_encoder", False),
+            "encoder_ckpt": config.get("encoder_ckpt"),
+        }
+        if encoder_type in {"slae", "esm"}:
+            encoder_config["embedding_key"] = "embedding"
+            encoder_config["embedding_dim"] = config.get("embedding_dim")
+
+    encoder = build_encoder(encoder_config, device)
```
DorisMai: verbatim to `inference.py` line 245-264. consider extract and import.

scripts/train_confidence.py
```diff
@@ -0,0 +1,815 @@
...
+        # partial window still steps; the stepping backward runs outside no_sync
+        # and all-reduces the whole accumulated gradient.
+        is_boundary = (micro % accum == 0) or (micro == n_batches)
+        sync_ctx = (
+            no_sync() if (no_sync is not None and not is_boundary) else nullcontext()
+        )
+
+        with sync_ctx:
+            if is_empty:
+                loss = sum(p.sum() for p in params) * 0.0
```
DorisMai: I don't have experience with DDP but Claude raised that DDP hangs if some rank does not have forward run (opposite from your comment in line 476).  I could not get a pod with GPU right now, so please verify and edit if necessary. Consider add a test.

scripts/train_confidence.py
```diff
@@ -0,0 +1,815 @@
...
+    state = torch.load(ckpt_path, map_location=device, weights_only=False)
+    if isinstance(state, dict) and "model_state_dict" in state:
+        state = state["model_state_dict"]
+
+    target = model.state_dict()
+    compatible = {
+        k: v for k, v in state.items() if k in target and target[k].shape == v.shape
+    }
+    missing = model.load_state_dict(compatible, strict=False).missing_keys
+    logger.info(
```
DorisMai, line 357: add a warning or raise if empty `compatible` in case `ckpt_path` is invalid

README.md
```diff
@@ -319,6 +321,31 @@ uv run torchrun --nproc_per_node=4 -m scripts.train \
...
 ```
 
+### Confidence Model Training
+
+Trains `ConfidenceGVP` to score flow-sampled candidate waters, reusing the flow
+run's cache layout and config plus a per-PDB candidate directory:
+
+```bash
+uv run python -m scripts.train_confidence \
+    --flow_run_dir <flow_run> \
```
DorisMai, line 332: missing required args `run_name`, which also doesn't have a help str in parse_args() on how it is used

src/utils.py
```diff
@@ -660,3 +660,34 @@ def save_protein_plot(
     ax.set_title(f"Step {step}")
     plt.savefig(f"{save_dir}/step_{step}.png")
     plt.close()
+
+
+def auc_pr_and_best_f1(scores: Tensor, labels: Tensor) -> tuple[float, float]:
```
DorisMai, line 665: why not use sklearn? you already compute another auc_pr in utils.py using sklearn. On that note, sklearn is never explicitly declared as dependency but seems somehow available through another dependency -- worth adding it explicitly. If you don't want to mess with this function here, maybe could add a test against sklearn's result?

scripts/train_confidence.py
```diff
@@ -0,0 +1,815 @@
...
+        encoder=encoder,
+        hidden_dims=(config.get("hidden_s") or 256, config.get("hidden_v") or 64),
+        edge_scalar_dim=config.get("edge_scalar_dim") or NUM_RBF,
+        layers=config.get("flow_layers") or 3,
+        drop_rate=config.get("drop_rate", 0.1),
+        n_message_gvps=config.get("n_message_gvps", 2),
+        n_update_gvps=config.get("n_update_gvps", 2),
+        cutoff=config.get("cutoff", 8.0),
+        max_neighbors=config.get("max_neighbors", 256),
+        dynamic_edge_policy="knn_if_isolated",
```
DorisMai, line 319: no passing of other config args like `knn_fallback_k`?

scripts/train_confidence.py
```diff
@@ -0,0 +1,815 @@
...
+                    "val/mae": val_metrics["mae"],
+                    "val/auc_pr": val_metrics["auc_pr"],
+                    "val/best_f1": val_metrics["best_f1"],
+                    "lr": lr,
+                },
+                step=step_counter,
+            )
+
+        aucpr = val_metrics["auc_pr"]
+        is_best = aucpr == aucpr and aucpr > best_aucpr  # `== aucpr` rejects nan
```
DorisMai: I've never seen `==` as a way to check nan...is this standard???

scripts/train_confidence.py
```diff
@@ -0,0 +1,815 @@
...
+            no_sync() if (no_sync is not None and not is_boundary) else nullcontext()
+        )
+
+        with sync_ctx:
+            if is_empty:
+                loss = sum(p.sum() for p in params) * 0.0
+            else:
+                preds = model(batch, return_logits=True)
+                loss = F.binary_cross_entropy_with_logits(preds, target)
+            (loss / accum).backward()
```
DorisMai, line 502: `/ accum` assumes every `micro` has the same size, is this true? like same n candidate same n structure?

DorisMai (approved):
there are a few questions I'd like to get clarifications on

scripts/train_confidence.py
vratins: Done, moved the block to src/encoder_base to call it.

scripts/train_confidence.py
vratins: Good catch, tried this with 2 GPUs. Rare occurrence in our case, but important to except out regardless. Fixed.

scripts/train_confidence.py
vratins: Done

README.md
vratins: Done

src/utils.py
vratins: a) added it explicitly, b) just to make things simple I do not use sklearn since I would need to call additional sklearn methods, and I keep it torch native.

scripts/train_confidence.py
vratins: Switched to `not math.isnan(aucpr)`

scripts/train_confidence.py
vratins: Added documentation to explain this but basically it assumes equal candidate counts per micro-batch.

scripts/train_confidence.py
vratins: Addressed

PR #99, opened 2026-08-18
I/O layer to merge predicted waters into a structure and write out PDB/CIF

vratins:
- Adds `src/structure_io.py`, a small structure-I/O layer that takes model output and converts into a writable structure for biotite. 
- end-to-end pipeline scripts will use these methods to write out structure files after inference. 

## Summary by CodeRabbit

* **New Features**
  * Added support for combining predicted water positions with existing molecular structures.
  * Automatically assigns water atoms, chains, residue numbers, occupancy, and B-factors.
  * Added structure export in PDB or CIF format based on the output file extension.

* **Bug Fixes**
  * Improved handling of tensor and array coordinate inputs, empty predictions, and existing structure annotations.

### DIFF
--- a/src/structure_io.py
+++ b/src/structure_io.py
@@ -1,0 +1,162 @@
+# structure_io.py
+
+"""Merge predicted waters with the protein+het atoms and write the result to PDB or CIF."""
+
+from __future__ import annotations
+
+import biotite.structure as bts
+import numpy as np
+from biotite.structure.io.pdb import PDBFile
+from biotite.structure.io.pdbx import CIFFile, set_structure as _set_structure_cif
+
+
+# Columns every AtomArray has. Any other column on the kept atoms (e.g. b_factor,
+# occupancy) must be mirrored onto the waters so the two arrays concatenate.
+_MANDATORY = (
+    "chain_id",
+    "res_id",
+    "ins_code",
+    "res_name",
+    "hetero",
+    "atom_name",
+    "element",
+)
+
+# Default water B-factor: mean B of kept atoms within this radius (A), plus an
+# offset since ordered waters move more than the atoms they touch.
+_B_FACTOR_CONTACT_RADIUS = 5.0
+_B_FACTOR_WATER_OFFSET = 10.0
+
+
+def merge_waters(
+    atoms: bts.AtomArray,
+    positions,
+    *,
+    chain_id: str | None = None,
+    b_factor: float | None = None,
+    occupancy: float = 1.0,
+) -> bts.AtomArray:
+    """Return the kept atoms with predicted waters appended as HOH oxygens.
+
+    Each row of positions becomes one O atom in its own HOH residue, in whatever
+    frame positions are given (no re-centering here).
+
+    Args:
+        atoms: Non-water atoms to keep (protein + hets), in the output frame.
+        positions: (N, 3) water coordinates; numpy array or tensor.
+        chain_id: Chain for the waters. Defaults to an unused single character.
+        b_factor: One B-factor for every water. Defaults to a per-water estimate
+            from the nearby atoms (see _default_b_factors). Written only if atoms
+            has a b_factor column.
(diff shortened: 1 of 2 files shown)


src/structure_io.py
```diff
@@ -0,0 +1,162 @@
...
+    return waters
+
+
+def _pick_unused_chain_id(atoms: bts.AtomArray) -> str:
+    """A single-character chain id not used by atoms (falls back to 'W')."""
+    used = set(atoms.chain_id.tolist()) if atoms.array_length() else set()
+    for c in "WXYZUVTSRQ0123456789":
+        if c not in used:
+            return c
+    return "W"
```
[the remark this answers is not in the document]
DorisMai: @vratins should raise

src/structure_io.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def _default_b_factors(atoms: bts.AtomArray, coords: np.ndarray) -> np.ndarray:
+    """Per-water default B-factor from the local environment.
+
+    Each water takes the mean B-factor of the kept atoms near it, plus an offset.
+    A water with no atom in range uses the overall mean. Returns zeros when atoms
+    has no b_factor.
```
DorisMai: add docstring on the arguments

src/structure_io.py
```diff
@@ -0,0 +1,162 @@
...
+    "res_name",
+    "hetero",
+    "atom_name",
+    "element",
+)
+
+# Default water B-factor: mean B of kept atoms within this radius (A), plus an
+# offset since ordered waters move more than the atoms they touch.
+_B_FACTOR_CONTACT_RADIUS = 5.0
+_B_FACTOR_WATER_OFFSET = 10.0
```
DorisMai, line 41: seems quite arbitrary...any support on why 10.0?

src/structure_io.py
```diff
@@ -0,0 +1,162 @@
...
+            occupancy column.
+
+    Returns:
+        The kept atoms followed by the waters, as one AtomArray.
+    """
+    coords = _to_coords(positions)
+    if chain_id is None:
+        chain_id = _pick_unused_chain_id(atoms)
+    if b_factor is None:
+        b_factor = _default_b_factors(atoms, coords)
```
DorisMai: you can probably move this into `_water_array` at line 120 since if the template doesn't have b_factor annotation there is no need to compute this. that could then simplify line 146 to only check array length.

src/structure_io.py
```diff
@@ -0,0 +1,162 @@
...
+        coords,
+        chain_id=chain_id,
+        b_factor=b_factor,
+        occupancy=occupancy,
+        template=atoms,
+    )
+    return atoms + waters
+
+
+def write_structure(atoms: bts.AtomArray, output_path: str) -> None:
```
DorisMai: do you care about preserving symmetry information? like unit cell and space group? If so, please add a real structure test (and both .cif and .pdb if you support both formats) to check this metadata is preserved, and maybe also check that the protein+water sits in the correct frame (like not zero centered or arbitrary translation/rotation etc.)

DorisMai (approved):
main question on metadata/frame, other things should be easy to address

src/structure_io.py
vratins: Arbitrary hyperparameter initially suggested by Stephanie, but supported by data on a couple of structures, on average the B-factor is higher by 10 compared to neighbouring atoms.

src/structure_io.py
vratins: Good point, will do.

src/structure_io.py
vratins: Done

src/structure_io.py
vratins: Done

src/structure_io.py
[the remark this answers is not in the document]
vratins: Done

PR #100, opened 2026-08-19
Building the flow-model input graph from a raw structure at inference time

vratins:
- The dataset class in `src` is currently wired for training and required ground truth waters (and associated filters) -- during inference we have access to no waters, just the protein. 
- Hence, adding `src/inference_graph.py`, where `build_inference_graph` turns a raw PDB/CIF into the exact HeteroData the flow model samples from — protein + optional symmetry mates + hets, waters dropped (if they exist in the cif file provided), coordinates centered on the ASU protein centroid.

## Summary by CodeRabbit

* **New Features**
  * Added support for generating inference-ready molecular graphs from PDB and CIF structures.
  * Structures are automatically centered, water molecules are excluded, and protein, ligand, and optional symmetry-mate information is included.
  * Graphs include spatial connections and cached model embeddings for inference.
  * Added configurable distance limits, neighbor limits, embedding selection, and optional graph persistence.
* **Tests**
  * Added coverage for structure formats, graph batching, ligand and symmetry handling, embedding requirements, and saved outputs.


src/inference_graph.py
```diff
@@ -0,0 +1,431 @@
...
+        out_dir: If given, the graph is also saved to out_dir/<cache_key>.pt.
+
+    Returns:
+        HeteroData with centred protein nodes (+ optional embeddings), empty
+        water nodes, and cached PP edges.
+    """
+    struc_path = str(struc_path)
+    if cache_key is None:
+        cache_key = Path(struc_path).stem
+    if encoder_type in ("slae", "esm") and processed_dir is None:
```
DorisMai, line 88: should guard against value outside of gvp slae and esm

tests/test_inference_graph.py
DorisMai, line 1: `test_protein_node_contract` and `test_ligands_appended_after_protein` are not needed, as `TestEquivalenceWithDataset` should already cover them. It would be better to parameterize equivalence test over `include_mates` and `include_ligands`.

DorisMai (approved):
- I understand a full refactor is not feasible now, but some new helper functions you add here should be directly reusable in `dataset.py`, such as `_attach_embeddings`, `residue_indexing`, and `_pp_edges`. Try to reuse them if you can.
- Make an issue that this script in the future should be further refactored
- In the context of future refactoring / maintenance, try to keep tests essential

src/inference_graph.py
vratins: Done

tests/test_inference_graph.py
vratins: Done

PR #101, opened 2026-08-20
bf16 AMP + fused AdamW for flow and confidence trainers

vratins:
Adding mixed precision (bfloat16) toggles in all training scripts.

## Summary by CodeRabbit

* **New Features**
  * Added optional CUDA bfloat16 automatic mixed precision for training and validation, enabled by default.
  * Added an option to enable fused AdamW optimization on CUDA-compatible hardware.
  * Added command-line controls to disable mixed precision when needed.

* **Bug Fixes**
  * Training automatically falls back to standard precision and optimizer behavior on non-CUDA devices.
  * Loss calculations remain stable by using full precision for final comparisons.

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -1146,9 +1146,7 @@ def main():
 
     # Resolve AMP once: it needs a CUDA device with bf16 support. Write the
     # verdict back to args so the recorded config.json reflects what actually ran.
-    if args.use_amp and (
-        device.type != "cuda" or not torch.cuda.is_bf16_supported()
-    ):
+    if args.use_amp and (device.type != "cuda" or not torch.cuda.is_bf16_supported()):
         reason = (
             "device is not CUDA"
             if device.type != "cuda"
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -1144,8 +1144,18 @@ def main():
     else:
         device = torch.device(args.device if torch.cuda.is_available() else "cpu")
 
-    if args.use_amp and device.type != "cuda":
-        logger.warning("--use_amp set but device is not CUDA; training without AMP.")
+    # Resolve AMP once: it needs a CUDA device with bf16 support. Write the
+    # verdict back to args so the recorded config.json reflects what actually ran.
+    if args.use_amp and (
+        device.type != "cuda" or not torch.cuda.is_bf16_supported()
+    ):
+        reason = (
+            "device is not CUDA"
+            if device.type != "cuda"
+            else "the GPU lacks bfloat16 support"
+        )
+        logger.warning(f"--use_amp set but {reason}; training without AMP.")
+        args.use_amp = False
 
     if args.run_name is None:
         args.run_name = generate_run_name(args)
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -385,6 +385,20 @@ def parse_args():
     )
     p.add_argument("--step_gamma", type=float, default=0.5, help="StepLR gamma")
 
+    # mixed precision / optimizer
+    p.add_argument(
+        "--use_amp",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+        help="Run the training forward pass under bfloat16 autocast (CUDA only). "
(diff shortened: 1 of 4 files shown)


src/flow.py
```diff
@@ -1122,16 +1134,17 @@ def training_step(
...
+            v_pred = self.model(batch, t)
 
         # target velocity
         v_target = x1_star - x0_star
 
-        # weighted MSE loss (upweight near t=1)
+        # weighted MSE loss (upweight near t=1), reduced in fp32
         w = 1.0 / (self.loss_eps + (1.0 - t_per_atom))
-        per_atom_mse = (v_pred - v_target).pow(2).mean(dim=-1, keepdim=True)
+        per_atom_mse = (v_pred.float() - v_target).pow(2).mean(dim=-1, keepdim=True)
```
DorisMai: would you need to cast it again for the rmsd calculation below? and same thing in validation (line 1220-1224)

scripts/train.py
```diff
@@ -1130,6 +1144,9 @@ def main():
     else:
         device = torch.device(args.device if torch.cuda.is_available() else "cpu")
 
+    if args.use_amp and device.type != "cuda":
```
DorisMai: why not set `args.use_amp=False` here? does config json save this? same in train_confidency.py

scripts/train_confidence.py
```diff
@@ -470,6 +482,7 @@ def train_one_epoch(
     total_loss, total_n = 0.0, 0
     params = [p for group in optimizer.param_groups for p in group["params"]]
     accum = max(1, args.grad_accum_steps)
+    use_amp = getattr(args, "use_amp", True) and device.type == "cuda"
```
DorisMai: why not use `args.use_amp`? same in line 574

DorisMai (approved):
I don't have direct experience with amp to review this properly, but try to look at coderabbitai's comments

scripts/train_confidence.py
```diff
@@ -565,8 +582,9 @@ def validate(
         target = batch["water"].target_confidence
         if target.numel() == 0:
             continue
-        preds = model(batch, return_logits=True)
-        loss = F.binary_cross_entropy_with_logits(preds, target)
+        with torch.autocast(device_type="cuda", dtype=torch.bfloat16, enabled=use_amp):
+            preds = model(batch, return_logits=True)
+            loss = F.binary_cross_entropy_with_logits(preds, target)
```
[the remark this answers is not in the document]
vratins: Yep makes sense, casting once right after autocast

scripts/train.py
vratins: That's fair, fixed it for both train scripts.

scripts/train_confidence.py
vratins: Done

src/flow.py
```diff
@@ -935,6 +939,14 @@ def __init__(
             self.graph_cutoff = DEFAULT_EDGE_CUTOFF
         self.sampling_strategy = sampling_strategy
 
+    def _autocast_context(self, device: torch.device):
+        """bf16 autocast on CUDA when --use_amp is set; a no-op otherwise."""
+        return torch.autocast(
+            device_type="cuda",
+            dtype=torch.bfloat16,
+            enabled=self.use_amp and device.type == "cuda",
+        )
```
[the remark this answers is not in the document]
vratins: Added a guard.

src/flow.py
vratins: Yep fair, casting it once and re-using.

vratins: As far as the review is concerned, torch does abstract a lot of the amp usage when using bfloat16 in the latest versions, so less likely to error out compared to when we needed context managers.

src/flow.py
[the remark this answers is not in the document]
vratins: Check again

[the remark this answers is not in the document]
vratins: done

PR #102, opened 2026-08-20
Adding flow trainer flags

vratins:
Adding flags during training:

- seeds
- eval method + number of steps
- selection metric (f1 vs auc vs blend)
- lr decay parameters
- resume logic

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -1395,6 +1395,21 @@ def main():
         best_sel_score = float("-inf") if best_sel_score is None else best_sel_score
         logger.info(f"Resumed from {ckpt_path} at epoch {start_epoch}.")
 
+    # A generative selection metric reads the sampling eval, so best.pt can only
+    # update on eval epochs. If none fall in the remaining schedule, fall back to
+    # val_loss so best.pt still gets written.
+    selection_metric = args.selection_metric
+    remaining_epochs = range(start_epoch + 1, args.epochs + 1)
+    if selection_metric != "val_loss" and not any(
+        e % args.eval_every == 0 for e in remaining_epochs
+    ):
+        logger.warning(
+            f"--selection_metric {selection_metric} needs an eval epoch, but none "
+            f"fall in epochs {start_epoch + 1}..{args.epochs} at --eval_every "
+            f"{args.eval_every}; selecting best.pt on val_loss instead."
+        )
+        selection_metric = "val_loss"
+
     for epoch in range(start_epoch + 1, args.epochs + 1):
         # Without this every epoch replays the same shard order on every rank.
         if ddp_is_active():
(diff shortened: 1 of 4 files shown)


scripts/train.py
```diff
@@ -706,12 +759,12 @@ def run_eval_sampling(
         if graph["water"].num_nodes == 0:
             continue
 
-        out = flow_matcher.rk4_integrate(
+        # rk4 returns a per-frame trajectory (used for GIFs); euler does not.
```
DorisMai: why the distinction? maybe add a warning if user pass args to save gif

scripts/train.py
```diff
@@ -980,6 +1041,8 @@ def save_checkpoint(
     optimizer_step_count,
     path,
     best=False,
+    best_val_loss=None,
+    best_sel_score=None,
```
DorisMai, line 1050: should add argument docstring. also, should you also store the selection metric? otherwise if upon reload metric is different, comparison of best becomes wrong, and should either raise or reset to some default value

scripts/train.py
```diff
@@ -692,12 +737,20 @@ def run_eval_sampling(
         run_dir: Path to run directory for saving outputs
     """
     flow_matcher.model.eval()
+    # Fix the eval-sampling RNG so the per-epoch generative metrics are comparable
+    # across epochs (the water prior and integration both draw noise).
+    torch.manual_seed(args.val_seed)
```
DorisMai: If you run eval once, is this seed globally set such that any training step after this is affected? seems potentially problematic, and it would also ignore your `--seed -1` CLI arg.

scripts/train.py
```diff
@@ -1126,6 +1192,14 @@ def main():
     """Run the full training pipeline."""
     args = parse_args()
 
+    # Seed weight init and data shuffling when requested; left unseeded otherwise.
+    # Seeded by default (--seed 42) for reproducibility; a negative seed opts out.
+    if args.seed >= 0:
+        random.seed(args.seed)
+        np.random.seed(args.seed)
```
DorisMai, line 1215: but this is overwritten on line 1293 to val_seed?

scripts/train.py
```diff
@@ -1412,6 +1507,59 @@ def main():
...
+        elif eval_metrics:  # generative metric, defined only on eval epochs
+            if selection_metric == "blend":
+                raw = (
+                    0.85 * eval_metrics["eval/avg_f1"]
+                    + 0.15 * eval_metrics["eval/avg_auc_pr"]
+                )
+            else:
+                raw = eval_metrics[f"eval/avg_{selection_metric}"]
+            sel_history.append(raw)
+            sel = sum(sel_history[-3:]) / len(sel_history[-3:])  # rolling-3 smoothed
```
DorisMai: consider extract the hardcoded numbers as constants like BLEND_F1_WEIGHT and SEL_ROLLING_WINDOW, and update the helper str in argparse correspondingly. If you have justification for why these magic numbers that's even better.

DorisMai (approved):
mostly minor comments, a bit confusion on the seeds

scripts/train.py
vratins: Will address in a docstring. The 0.85 is really just an arbitrary hp.

scripts/train.py
vratins: Fixed.

scripts/train.py
vratins: Yes, fixed this.

scripts/train.py
vratins: Both done.

scripts/train.py
vratins: Honestly just a development thing, I stopped using gifs to view flow when I moved to euler, will adjust this to return gif in either case for consistency.

PR #105, opened 2026-08-20
End-to-end water prediction pipeline script

vratins:
- Adds `scripts/predict_waters.py`: predict waters for a raw PDB/CIF, end to end.
- Pipeline: Strip to protein + hets → sample candidates from a flow checkpoint → score with a confidence checkpoint → cluster (vdW) and threshold → return to the input frame → write.
- Output (per structure, in the input frame): a)`<name>_pred.{pdb,cif}` — input protein + hets with predicted waters as HOH oxygens b)`<name>_waters.txt` — `x y z confidence`, one row per water
- Mnor edit in `confidence.py` docstring to remove self-conditioning reference.


scripts/predict_waters.py
```diff
@@ -0,0 +1,381 @@
...
+from src.constants import NUM_RBF
+from src.dataset import element_onehot, parse_asu_with_biotite
+from src.encoder_base import build_encoder
+from src.flow import FlowMatcher
+from src.inference_graph import build_inference_graph
+from src.structure_io import merge_waters, write_structure
+from src.utils import setup_logging_for_tqdm
+
+
+VDW_RADIUS_A = 1.52  # oxygen vdW radius; clustering/NMS radius
```
DorisMai: should probably defined in `confidence.py` (which has a hardcoded 1.52 default) and this script (and any confidence related tests) import from there

scripts/predict_waters.py
```diff
@@ -0,0 +1,381 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Model loading
+# ---------------------------------------------------------------------------
+
+
+def build_confidence_model(config: dict, device: torch.device) -> ConfidenceGVP:
+    """Instantiate ConfidenceGVP from a flow run's config (mirrors train_confidence)."""
```
DorisMai: why not import the same function from `train_confidence.py`? any particular difference here? If it is indeed re-usable, maybe should move this function to `confidence.py` and have both this and `train_confidence.py` import from there.

scripts/predict_waters.py
```diff
@@ -0,0 +1,381 @@
...
+        return conf_model(scored).detach().cpu()
+
+
+def select_waters(
+    candidate_pos: torch.Tensor,
+    confidences: torch.Tensor,
+    *,
+    mode: str,
+    threshold: float | None = None,
+    radius: float = VDW_RADIUS_A,
```
DorisMai: an argument never passed by any caller fyi

scripts/predict_waters.py
```diff
@@ -0,0 +1,381 @@
...
+    candidate_pos: torch.Tensor,
+    device: torch.device,
+) -> torch.Tensor:
+    """Confidence score in [0, 1] for each candidate, given the protein graph."""
+    n = candidate_pos.size(0)
+    if n == 0:
+        return candidate_pos.new_zeros(0)
+    scored = graph.clone()
+    scored["water"].pos = candidate_pos.to(device)
+    scored["water"].x = element_onehot(["O"] * n).to(device)
```
DorisMai: re-use `confidence_dataset._oxygen_features`

scripts/predict_waters.py
```diff
@@ -0,0 +1,381 @@
...
+        encoder=encoder,
+        hidden_dims=(config.get("hidden_s") or 256, config.get("hidden_v") or 64),
+        edge_scalar_dim=config.get("edge_scalar_dim") or NUM_RBF,
+        layers=config.get("flow_layers") or 3,
+        drop_rate=config.get("drop_rate", 0.1),
+        n_message_gvps=config.get("n_message_gvps", 2),
+        n_update_gvps=config.get("n_update_gvps", 2),
+        cutoff=config.get("cutoff", 8.0),
+        max_neighbors=config.get("max_neighbors", 256),
+        dynamic_edge_policy="knn_if_isolated",
```
DorisMai: no passing knn_fallback_k?

scripts/predict_waters.py
```diff
@@ -0,0 +1,381 @@
...
+        sel_pos, sel_conf = select_waters(
+            candidate_pos,
+            conf,
+            mode=args.selection,
+            threshold=args.threshold,
+        )
+        # Back to the input frame, then write structure + scored coordinates.
+        water_xyz = sel_pos.numpy() + center
+        structure = merge_waters(kept, water_xyz)
+        write_structure(structure, str(out_dir / f"{name}_pred{args.out_format}"))
```
DorisMai: no space group passed

scripts/predict_waters.py
```diff
@@ -0,0 +1,381 @@
...
+        )
+    raise ValueError(f"Unknown selection mode: {mode!r} (only 'confidence' so far)")
+
+
+# ---------------------------------------------------------------------------
+# Per-structure prediction
+# ---------------------------------------------------------------------------
+
+
+def _kept_atoms_and_center(struc_path: str):
```
DorisMai: add return type hint and docstring

scripts/predict_waters.py
```diff
@@ -0,0 +1,381 @@
...
+    """Protein + hets to write out, and the ASU protein centroid used to centre.
+
+    Recomputes the centroid build_inference_graph used, so predicted waters
+    (sampled in the centred frame) return to the input reference frame.
+    """
+    protein_atoms, _waters, ligand_atoms = parse_asu_with_biotite(struc_path)
+    center = protein_atoms.coord.mean(axis=0)
+    kept = protein_atoms + ligand_atoms if len(ligand_atoms) else protein_atoms
```
DorisMai, line 133: should `ligand_atoms` always be written out, even if `include_ligand=False`? I don't have a strong opinion on this, but worth a comment

tests/test_predict_waters.py
```diff
@@ -0,0 +1,155 @@
...
+    load_state_dict_lenient,
+    predict_structures,
+    select_waters,
+)
+from src.confidence import ConfidenceGVP
+from src.flow import FlowMatcher, FlowWaterGVP
+
+
+@pytest.mark.unit
+class TestSelectWaters:
```
DorisMai: seems unnecessary, essentially testing `cluster_waters_vdw`, covered in `test_confidence.py`

tests/test_predict_waters.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def test_pdb_list_warns_on_missing(self, tmp_path):
+        lst = tmp_path / "list.txt"
+        lst.write_text("does_not_exist\n")
+        paths = _collect_struc_paths(
+            SimpleNamespace(struc=None, pdb_list=str(lst), base_pdb_dir=str(tmp_path))
+        )
+        assert paths == []
+
+    def test_kept_atoms_drop_waters_and_center_is_protein_centroid(self, pdb_4h0b):
```
DorisMai: checking protein and ligand coords close would be stronger than just checking centroid. also should check metadata like box (unit cell), maybe symmetry if appropriate here

scripts/predict_waters.py
```diff
@@ -0,0 +1,381 @@
...
+    p.add_argument("--out_format", default=".pdb", choices=[".pdb", ".cif"])
+
+    p.add_argument(
+        "--selection",
+        default="confidence",
+        choices=["confidence"],
+        help="Final-water selection rule.",
+    )
+    p.add_argument(
+        "--threshold",
```
DorisMai: consider rename to `confidence_threshold`, as `train/inference.py` has its own `--threshold` on distance. And would be helpful to add a range validation

scripts/predict_waters.py
```diff
@@ -0,0 +1,381 @@
...
+
+"""End-to-end water prediction from a raw PDB/CIF file.
+
+Takes structures (with or without waters), strips them to protein + hets, samples
+candidate waters from a trained flow checkpoint, scores them with a confidence
+checkpoint, clusters and selects the final set, and writes out predicted-water
+structures (PDB/CIF) plus a text file of their coordinates and confidence.
+
+This is the *prediction* entry point; inference.py remains the cache-based
+evaluation/benchmark tool. Shared flow machinery is imported, not duplicated.
```
DorisMai: this docstring was a bit confusing to me. I think `inference.py` is essentially evaluation of the generator model, and maybe you would have another script `inference_confidence.py` that evaluates predicted water (generator + confidence) against GT water. This prediction script would essentially be mirroring the whole pipeline `inference_confidence.py` except that there is not GT. Is this understanding correct?

DorisMai (approved):
mostly minor refactoring comments

scripts/predict_waters.py
vratins: Done.

scripts/predict_waters.py
vratins: Done

scripts/predict_waters.py
vratins: Remvoed

scripts/predict_waters.py
vratins: Done

scripts/predict_waters.py
vratins: Fixed

scripts/predict_waters.py
vratins: Done

scripts/predict_waters.py
vratins: Added a comment for default behaviour for now

tests/test_predict_waters.py
vratins: Yep, cut it

scripts/predict_waters.py
vratins: Fixed

tests/test_predict_waters.py
vratins: Done

scripts/predict_waters.py
vratins: Yes, exactly, re-wrote the docstring to be clearer

scripts/predict_waters.py
vratins: done

PR #106, opened 2026-08-21
Add density selection mode to predict_waters

vratins:
Final major functional change to codebase:

Adds `--selection density` to `scripts/predict_waters.py` alongside the existing `confidence` mode.

- confidence (default): threshold candidates at `--confidence_threshold` (default 0.5), then cluster; every centroid is kept.
- density: cluster all candidates, then keep the top `floor(--density_ratio × ASU residues)` centroids by confidence (default ratio 0.6). No score cutoff, so the water count is fixed per structure.

## Summary by CodeRabbit

* **New Features**
  * Added density-based water selection alongside confidence-threshold selection.
  * Added command-line options for configuring selection mode and density parameters.
  * Added validation and sensible defaults for selection settings.

* **Bug Fixes**
  * Improved handling of invalid or incomplete selection options.
  * Ensured selection settings are consistently applied during predictions.


scripts/predict_waters.py
```diff
@@ -265,14 +286,22 @@ def parse_args() -> argparse.Namespace:
...
-        help="Drop candidates scoring below this, in [0, 1] (confidence mode).",
+        default=None,
+        help="confidence mode: drop candidates scoring below this, in [0, 1] "
+        f"(default {DEFAULT_CONFIDENCE_THRESHOLD}).",
+    )
+    p.add_argument(
+        "--density_ratio",
+        type=float,
+        default=None,
+        help="density mode: keep floor(ratio * ASU residues) waters by confidence "
```
DorisMai: I would prefer to be more exact on inclusive/exclusive (<= vs <) on these helper str. Also for the density mode, technically it would be min(top K, total) waters

scripts/predict_waters.py
```diff
@@ -293,11 +322,24 @@ def parse_args() -> argparse.Namespace:
...
-        p.error("--confidence_threshold must be in [0, 1]")
+    # Each mode has one knob. Reject the other mode's and fill the default.
+    if args.selection == "confidence":
+        if args.density_ratio is not None:
+            p.error("--density_ratio only applies to --selection density")
+        if args.confidence_threshold is None:
+            args.confidence_threshold = DEFAULT_CONFIDENCE_THRESHOLD
+        if not 0.0 <= args.confidence_threshold <= 1.0:
+            p.error("--confidence_threshold must be in [0, 1]")
+    else:
```
DorisMai: probably should validate selection mode here too, has to be either confidence or density

scripts/predict_waters.py
```diff
@@ -105,15 +112,27 @@ def select_waters(
...
         return cluster_waters_vdw(candidate_pos, confidences, threshold=threshold)
-    raise ValueError(f"Unknown selection mode: {mode!r} (only 'confidence' so far)")
+    if mode == "density":
+        if density_ratio is None or num_asu_residues is None:
+            raise ValueError(
+                "density selection needs density_ratio and num_asu_residues"
+            )
+        pos, conf = cluster_waters_vdw(candidate_pos, confidences)  # sorted desc
+        n_keep = int(density_ratio * num_asu_residues)
+        return pos[:n_keep], conf[:n_keep]
```
DorisMai, line 144: should probably warn if `n_keep > len(pos)`

tests/test_predict_waters.py
```diff
@@ -26,9 +27,64 @@
...
 @pytest.mark.unit
-def test_select_waters_unknown_mode_raises():
-    with pytest.raises(ValueError, match="mode"):
-        select_waters(torch.zeros(1, 3), torch.ones(1), mode="density")
+class TestSelectWaters:
+    def test_unknown_mode_raises(self):
+        with pytest.raises(ValueError, match="mode"):
+            select_waters(torch.zeros(1, 3), torch.ones(1), mode="bogus")
+
+    def test_density_keeps_top_n_by_confidence(self):
```
DorisMai, line 35: would be good to add the edge cases here like no waters survives the thresholding (e.g. confidence == 1, or density*num_asu_residue < 1) -- which probably also should have a warning in the production code too.

tests/test_predict_waters.py
```diff
@@ -113,6 +169,7 @@ def test_pipeline_writes_predicted_structure(self, pdb_4h0b, gvp_encoder, tmp_pa
             water_ratio=1.0,
             selection="confidence",
             confidence_threshold=0.0,  # keep all, so the write path is exercised
+            density_ratio=None,
```
DorisMai: why not parameterize the integration tests over [confidence, density]?

DorisMai (approved):
minor improvement suggestions

tests/test_predict_waters.py
vratins: Done

tests/test_predict_waters.py
vratins: Done

scripts/predict_waters.py
vratins: Added

scripts/predict_waters.py
vratins: I believe argparse choices already enforces this before we reach the branch; anything else exits with the usual usage error. Added a comment toward this instead

scripts/predict_waters.py
vratins: Done

scripts/predict_waters.py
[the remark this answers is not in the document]
vratins: Done

PR #107, opened 2026-08-22
Adding trained checkpoints

vratins:
Adding the trained checkpoints and their configs. Uploaded to git lfs.

## Summary by CodeRabbit

- **New Features**
  - Added pretrained flow and confidence model checkpoints for both mates-enabled and mates-disabled workflows.
  - Added configuration presets covering data preparation, model setup, training, evaluation, and checkpoint recovery.
- **Chores**
  - Improved handling of large checkpoint files through Git LFS, enabling efficient versioning and retrieval of model assets.
  - Included the required metadata so distributed checkpoint files can be restored reliably.


DorisMai (approved):
lgtm
- Might worth addressing coderabbitai's suggestion to remove/abstract diffuse paths. The original .json file you used should stay somewhere internally (gdrive or whatever). 
- I don't know if it is standard to have an integration test on real/released checkpoint, but seems tricky since esm embedding needed?
- nitpick: probably don't need to save optimizer/scheduler in checkpoint, just weights?

vratins:
Agreed, got rid of opt/sched in the checkpoint. Got rid of the paths in config.

PR #109, opened 2026-08-22
Minor edit to loss function

vratins:
We train the shipped checkpoints with uniform loss, and no weighting at any time points -- hence removing redundant code. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated velocity-error calculations to use consistent, unweighted mean squared error during training and validation.
  * Improved per-graph diagnostic loss reporting by averaging error across atoms.
  * Retained safeguards for detecting invalid training values such as NaNs.

### DIFF
--- a/src/flow.py
+++ b/src/flow.py
@@ -909,7 +909,6 @@ class FlowMatcher:
     def __init__(
         self,
         model,
-        loss_eps: float = 1e-3,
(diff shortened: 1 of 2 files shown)


DorisMai (approved):
lgtm. just double checking that other hyperparameters are not stale, like lr and grad clipping

PR #110, opened 2026-08-24
Cleaning up prediction script, and removing hard coded paths.

vratins:
Final clean up for the code to make it completely functional and runnable from the READMe (1dev_readme`). 

Comment clean up, hard coded path removal, setting names properly for the candidate caches to ensure right caches files are used if re-running inference. 

## Summary by CodeRabbit

- **New Features**
  - Generate embeddings from raw PDB/CIF files or split-file datasets.
  - Added geometry caching to speed up repeated water-prediction runs.
  - Checkpoint-based model loading now validates required parameters.
  - Training defaults now use ESM embeddings and configurable checkpoint and logging locations.

- **Bug Fixes**
  - Improved coordinate restoration using the protein graph centroid.
  - Added validation for required embeddings and command-line inputs.
  - Reused cached graphs and candidate waters for consistent, faster inference.

- **Documentation**
  - Updated training and inference examples with required dataset and checkpoint options.

### DIFF
--- a/scripts/generate_esm_embeddings.py
+++ b/scripts/generate_esm_embeddings.py
@@ -7,13 +7,13 @@ This script:
 3. Sanitizes the AtomArray in-memory (removes HETATM flags, standardizes residue
    names) so ESM3 structurally embeds non-canonical residues (mapped to 'UNK'/'X').
 4. Runs ESM3 to get per-residue embeddings.
-5. Saves embeddings to separate .pt files in cache_dir/esm/
+5. Saves embeddings to separate .pt files in processed_dir/esm/
 
 Usage:
     # training layout: keyed by split entry ('6eey_final' -> esm/6eey_final.pt)
     uv run python -m scripts.generate_esm_embeddings \
-        --split_file /path/to/split.txt \
-        --cache_dir /path/to/cache \
+        --pdb_list /path/to/split.txt \
+        --processed_dir /path/to/cache \
         --base_pdb_dir /path/to/pdbs \
         [--device cuda:0] \
         [--model_name esm3-open] \
@@ -24,7 +24,7 @@ Usage:
     # is how predict_waters looks embeddings up
     uv run python -m scripts.generate_esm_embeddings \
         --struc protein.cif other.pdb \
-        --cache_dir /path/to/cache
+        --processed_dir /path/to/cache
 
 Output format (per cache file):
     {
@@ -37,6 +37,7 @@ Output format (per cache file):
 
 import argparse
 import io
+from collections import Counter
 from pathlib import Path
 
 import numpy as np
@@ -170,33 +171,31 @@ def main() -> None:
     parser = argparse.ArgumentParser(
         description="Generate ESM3 embeddings for protein structures"
     )
-    parser.add_argument(
-        "--split_file",
+    src = parser.add_mutually_exclusive_group(required=True)
+    src.add_argument(
+        "--pdb_list",
         type=Path,
-        default=None,
         help="Text file with PDB entries (one per line, e.g., '6eey_final'); "
         "resolved under --base_pdb_dir and keyed by split entry.",
     )
-    parser.add_argument(
+    src.add_argument(
         "--struc",
         type=Path,
         nargs="+",
-        default=None,
         help="One or more raw PDB/CIF files, keyed by file stem "
-        "(protein.cif -> esm/protein.pt). Use this for prediction inputs. "
-        "Takes precedence over --split_file.",
+        "(protein.cif -> esm/protein.pt). Use this for prediction inputs.",
     )
     parser.add_argument(
-        "--cache_dir",
+        "--processed_dir",
         type=Path,
         required=True,
-        help="Base cache directory; embeddings saved to {cache_dir}/esm/",
+        help="Base cache directory; embeddings saved to {processed_dir}/esm/",
     )
     parser.add_argument(
         "--base_pdb_dir",
         type=Path,
         default=None,
-        help="Base directory containing PDB subdirectories; required with --split_file",
+        help="Base directory containing PDB subdirectories; required with --pdb_list",
     )
     parser.add_argument(
         "--model_name",
@@ -223,21 +222,21 @@ def main() -> None:
     )
 
     args = parser.parse_args()
-    if not args.struc and not args.split_file:
-        parser.error("provide --struc <files> or --split_file <file>")
-    if args.split_file and args.base_pdb_dir is None:
-        parser.error("--split_file requires --base_pdb_dir")
+    if not args.struc and not args.pdb_list:
+        parser.error("provide --struc <files> or --pdb_list <file>")
+    if args.pdb_list and args.base_pdb_dir is None:
+        parser.error("--pdb_list requires --base_pdb_dir")
+    if args.struc:
+        counts = Counter(p.stem for p in args.struc)
+        duplicates = sorted(s for s, n in counts.items() if n > 1)
+        if duplicates:
+            parser.error(
+                f"duplicate structure names {duplicates}: embeddings are keyed "
+                "by file stem, so every --struc file needs a distinct name"
+            )
 
-    esm_cache_dir = args.cache_dir / "esm"
+    esm_cache_dir = args.processed_dir / "esm"
     esm_cache_dir.mkdir(parents=True, exist_ok=True)
-
-    device = torch.device(args.device if torch.cuda.is_available() else "cpu")
-    logger.info(f"Using device: {device}")
-
-    logger.info(f"Loading ESM3 model: {args.model_name}...")
-    model = ESM3.from_pretrained(args.model_name).to(device)
-    model.eval()
-    logger.info("Model loaded.")
 
     if args.struc:
         # Key by file stem, matching how predict_waters looks embeddings up.
@@ -246,7 +245,7 @@ def main() -> None:
         ]
         logger.info(f"Embedding {len(entries)} structure(s) passed via --struc")
     else:
-        entries = parse_split_file(args.split_file, args.base_pdb_dir)
+        entries = parse_split_file(args.pdb_list, args.base_pdb_dir)
         logger.info(f"Found {len(entries)} entries in split file")
 
     if args.batch_limit:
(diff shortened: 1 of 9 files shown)


scripts/generate_esm_embeddings.py
```diff
@@ -19,6 +20,12 @@
         [--batch_limit 10] \
         [--overwrite]
 
+    # raw structures: keyed by file stem (protein.cif -> esm/protein.pt), which
+    # is how predict_waters looks embeddings up
+    uv run python -m scripts.generate_esm_embeddings \
+        --struc protein.cif other.pdb \
+        --cache_dir /path/to/cache
```
DorisMai: `cache_dir` here (and generate slae) vs `processed_dir` in `predict_waters.py` (`inference.py`, and maybe other scripts) are essentially the same thing? Consider renaming all occurences to be consistent.

scripts/predict_waters.py
```diff
@@ -292,6 +354,13 @@ def parse_args(argv: list[str] | None = None) -> argparse.Namespace:
         "generate_slae_embeddings.py first. Looked up by file stem under "
         "processed_dir/<encoder_type>.",
     )
+    p.add_argument(
+        "--geometry_cache",
```
DorisMai: this is a confusing name...it's not just graph, but also candidate waters. consider rename to something like predict_cache

scripts/predict_waters.py
```diff
@@ -171,40 +174,82 @@ def predict_structures(
     args: argparse.Namespace,
     device: torch.device,
 ) -> None:
-    """Predict + write final waters for a batch of structures."""
+    """Predict + write final waters for a batch of structures.
+
+    With --geometry_cache set, inference graphs are cached at
+    <geometry_cache>/<name>[_mates].pt and candidate waters at
+    <geometry_cache>/candidates/<name>[_mates]_<ckpt>_<method><steps>_r<ratio>.pt,
+    and both are reused when present.
```
DorisMai: In this case, you could have avoid loading the flow model checkpoint, checking esm embeddings, and ignore a bunch of other flow sampling related cli args. consider improve the arg parse function given this.

scripts/predict_waters.py
```diff
@@ -171,40 +174,82 @@ def predict_structures(
...
+                encoder_type=encoder_type,
+                processed_dir=args.processed_dir,
+                include_mates=args.include_mates,
+                include_ligands=flow_config.get("include_ligands", True),
+                cutoff=flow_config.get("cutoff", 8.0),
+                max_neighbors=flow_config.get("max_neighbors", 256),
+            )
+            if graph_pt is not None:
+                cache.mkdir(parents=True, exist_ok=True)
+                torch.save(graph, graph_pt)
```
DorisMai: nitpick: could have absorb this into `build_inference_graph` which only enables saving but under a different name

scripts/train.py
```diff
@@ -225,7 +220,12 @@ def parse_args():
 
     # model
     p.add_argument(
-        "--encoder_type", type=str, default="gvp", choices=["gvp", "slae", "esm"]
+        "--encoder_type",
+        type=str,
+        default="esm",
+        choices=["gvp", "slae", "esm"],
+        help="Protein encoder. 'esm' (default) and 'slae' need embeddings "
```
DorisMai, line 227: add 'slae' (obsolete) or something like that indicate no longer supported (unless it works?)

scripts/predict_waters.py
```diff
@@ -265,16 +308,35 @@ def _collect_struc_paths(args: argparse.Namespace) -> list[str]:
...
+    (<processed_dir>/<encoder_type>/<stem>.pt) is missing. gvp needs none.
+    """
+    if encoder_type == "gvp":
+        return
+    if processed_dir is None:
+        raise SystemExit(f"--processed_dir is required for encoder_type={encoder_type}")
+    emb_dir = Path(processed_dir) / encoder_type
+    missing = [p for p in paths if not (emb_dir / f"{Path(p).stem}.pt").exists()]
+    if missing:
+        raise SystemExit(
```
DorisMai: why SystemExit instead of ValueError? same for line 320 above

tests/test_predict_waters.py
```diff
@@ -165,13 +191,39 @@ def test_pdb_list_warns_on_missing(self, tmp_path):
...
+class TestCheckEmbeddings:
+    def test_gvp_skips_check(self):
+        # gvp needs no embeddings, so nothing is required even with no cache.
+        _check_embeddings(["/no/such/protein.cif"], "gvp", None)
+
+    def test_missing_esm_embedding_raises_naming_files(self, tmp_path):
+        with pytest.raises(SystemExit) as exc:
+            _check_embeddings(["a/protein.cif", "b/other.pdb"], "esm", str(tmp_path))
+        msg = str(exc.value)
+        assert "protein.cif" in msg and "other.pdb" in msg
```
DorisMai: nitpick: could probably do `match="..."` with regex like elsewhere in this code base

tests/test_train_config.py
DorisMai, line 210: should have match message checking the exact kind of error tested here

tests/test_train_config.py
```diff
@@ -212,7 +216,18 @@ def test_dataset_defaults_match_train_defaults(monkeypatch):
     from src.dataset import ProteinWaterDataset
 
     monkeypatch.setattr(
-        "sys.argv", ["train.py", "--train_list", "t.txt", "--val_list", "v.txt"]
+        "sys.argv",
```
DorisMai, line 221: you should probably check over all eligible args, and it looks like there will be a few things that no longer matches between dataset and train:
- encoder_type
- include_mates
- cutoff (coincidental matching, not enforced)
- max_neighbor (coincidental matching, not enforced)

scripts/inference.py
```diff
@@ -7,9 +7,11 @@
 and outputs plots, gifs, and metrics for each PDB.
 
 Usage:
-    python scripts/inference.py \
+    python -m scripts.inference \
         --run_dir /path/to/run_directory \
```
DorisMai: I would suggest rename run_dir to flow_run_dir to match train_confidence.py

DorisMai (changes requested):
My biggest complaint is the various cli arg names across scripts that describe the same thing. I had an inline comment about cache_dir vs processed_dir, but there are also other things like:
- “geometry cache” appears in different arg names and can mean different things. geometry_cache_name from train.py and geometry_cache from inference.py seems to be the same thing? But geometry_cache in predict_waters.py refers to something different. Please clarify, make train/inference consistent and rename in predict_waters.py.
- Minor inconsistent things like 
    - struc is N structures in generate_esm but 1 structure in predict_waters.py
    - split_file vs pdb_list
    - out_dir vs output_dir

If fixing the cache_dir and geometry_cache are not feasible, make it explicit that "X flag here is the same (or NOT THE SAME) as Y flag there", or have a table of these corresponding folder/file path args in README.

scripts/train.py
vratins: Added a comment

scripts/generate_esm_embeddings.py
```diff
@@ -206,20 +222,31 @@ def main() -> None:
     )
 
     args = parser.parse_args()
+    if not args.struc and not args.pdb_list:
```
DorisMai: technically no longer needed since they are in required but mutually exclusive grup

scripts/predict_waters.py
```diff
@@ -240,44 +281,74 @@ def predict_structures(
...
+    else:
+        base = Path(args.base_pdb_dir)
+        names = [
+            ln.strip()
+            for ln in Path(args.pdb_list).read_text().splitlines()
+            if ln.strip()
+        ]
+        paths = []
+        for name in names:
+            if (base / name).suffix.lower() in (".cif", ".pdb"):
```
DorisMai, line 301: I just realized that the expected folder layout is different. Here it would be `base / structure.cif`, but all other scripts it would be `base / pdbid / pdbid_final.cif`. Maybe I shouldn't have requested to name this cli args to be the same...sorry...maybe specify this layout in the CLI arg helper string in this script and/or revert back to different cli arg names then separately document the layout in README.

scripts/predict_waters.py
```diff
@@ -361,29 +439,54 @@ def main() -> None:
...
+    # different settings never share one (the cache_candidates scheme). The
+    # checkpoint is identified by its directory name, so retraining in place
+    # makes the cache stale.
+    cache = None
+    if args.predict_cache:
+        mates = "mates" if args.include_mates else "nomates"
+        cache = Path(args.predict_cache) / (
+            f"{ckpt_dir.resolve().name}_{mates}"
+            f"_{args.method}{args.num_steps}_r{args.water_ratio:g}"
+        )
```
DorisMai, line 477: Probably should warn if the `include_mates` from `flow_config` is inconsistent with the `args.include_mates` here. also, the cache name can be a bit weird: you would get `mates_mates_<method><steps>_r<ratio>` or `mates_off_nomates_<method><steps>_r<ratio>` using the checkpoints from you. Maybe add a separating word in between like `<ckpt_name>_ckpt_<mates>_<method><steps>_r<ratio>`

DorisMai (approved):
Just minor comments for this PR left. However, Claude made a good point that for a user that use `predict_waters.py` with default args (which is mates ON checkpoint) on a predicted protein structure, there is currently no warning/error raised for either (1) the cif/pdb has no symmetry, or (2) AF DB structure that carries dummy symmetry like unit cell (1,1,1,90,90,90) with P1 symmetry. Probably should confirm/address this in separate PR, maybe not necessarily in `predict_waters.py`, could be in `dataset.py`.

scripts/predict_waters.py
vratins: I kept it separate so that the user does not need to follow a strict layout and can just drop files into a directory, but it would work either way -- but yes I think having the same name might clash when training

scripts/generate_esm_embeddings.py
vratins: Done

scripts/predict_waters.py
vratins: Done

PR #115, opened 2026-08-26
Switch to mates-off model if symmetry information is missing/incorrect

### LINKED ISSUE #114: Adding checks in the prediction script to identify whether the crystal information exists/is valid.
Based on @DorisMai comment in PR #110, where with default args (which is mates ON checkpoint) on a predicted protein structure, there is currently no warning/error raised for either (1) the cif/pdb has no symmetry, or (2) AF DB structure that carries dummy symmetry like unit cell (1,1,1,90,90,90) with P1 symmetry

vratins:
Currently if the symmetry information is missing, the model predicts with the mates on model, just without mates nodes as context. Worse, if the file has bad symmetry (like placeholders), we get spurious mates nodes.

This PR adds a check for the symmetry mates quality and routes to the mates off model if the check fails. Addressing issue #114.

### DIFF
--- a/scripts/predict_waters.py
+++ b/scripts/predict_waters.py
@@ -16,8 +16,9 @@ For esm/slae encoders the protein embeddings must already be in
 
 Models are loaded from a --ckpt_dir holding flow.pt, confidence.pt,
 flow_config.json and confidence_config.json. It defaults to the mates models
-shipped in the repo (checkpoints/mates); pass checkpoints/mates_off to run
-without symmetry mates.
+shipped in the repo (checkpoints/mates), or to checkpoints/mates_off when no
+input has usable crystal symmetry (predicted structures); pass a directory to
+choose one yourself.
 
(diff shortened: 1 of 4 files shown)


scripts/predict_waters.py
```diff
@@ -317,6 +315,43 @@ def _collect_struc_paths(args: argparse.Namespace) -> list[str]:
     return paths
 
 
+def resolve_ckpt_dir(explicit: str | None, paths: list[str]) -> Path:
+    """Pick the checkpoint directory when the user did not name one.
+
+    The shipped mates models expect crystal symmetry in the input. When no
```
marcuscollins, line 321: You aren't shipping models in the repo are you? I think it is generally a bad idea to do that.

marcuscollins (approved):

scripts/predict_waters.py
vratins: Addressed with git lfs

